PR #65, opened 2022-01-10
Hw test regular

maxkmaxk:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,11 +1,38 @@
 package core.basesyntax.service;
 
+import core.basesyntax.db.Storage;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
```
MaliukDaria, line 6: lets not use Storage directly, we have StorageDao for this purpose

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,188 @@
...
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.MethodOrderer.OrderAnnotation;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
+
+/**
+ * Feel free to remove this class and create your own.
+ */
+@TestMethodOrder(OrderAnnotation.class)
```
MaliukDaria, line 23: When we write unit tests they should not depend on the order of execution, lets remove execution order

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,188 @@
...
+    private static final String loginDefault = "login";
+    private static final String passNull = null;
+    private static final String passShort = "short";
+    private static final String passDefault = "password";
+    private static final Integer ageNull = null;
+    private static final Integer ageYoung = 17;
+    private static final Integer ageBarely = 18;
+    private static final Integer ageMax = Integer.MAX_VALUE;
+    private static final Integer ageMin = Integer.MIN_VALUE;
+    private static final Integer ageDefault = 25;
```
MaliukDaria, line 36: I dont think we need this variables at class level, and we really don't need any individual variables for null values

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,188 @@
...
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.people.clear();
+    }
+
+    @Test
+    @DisplayName("Correct login handling")
```
MaliukDaria, line 50:
```suggestion
    @DisplayName("register valid user ok")
```

Let's stick to this style: "There are a lot of ways to name your test methods. The main point is that they should have informative  names and be consistent along with other developers in your team. For this task use such convention: `<methodUnderTest>_<state>_<expectedBehavior>`; For example, if we are testing the method `register` with a `null`  user's age the test method name should be `register_nullAge_notOk`. notOk is because the test expects the register method to throw an exception."

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,188 @@
...
+    void register_login_Ok() {
+        User expected = getUser(id, loginDefault, passDefault, ageDefault);
+        User actual = registrationService.register(expected);
+        assertEquals(expected, actual, "register method return value");
+        assertEquals(1, Storage.people.size(), "Storage size");
+        assertEquals(expected, Storage.people.get(0), "added to storage user");
+    }
+
+    @Test
+    @DisplayName("Null login handling")
```
MaliukDaria, line 61:
```suggestion
    @DisplayName("register null login not ok")
```

MaliukDaria (changes requested):

MaliukDaria (changes requested):
test

PR #75, opened 2022-01-30
First raw version 77%,

manglhorn:
какие еще проверки можно устроить?

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -6,6 +6,16 @@ public class RegistrationServiceImpl implements RegistrationService {
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null || user.getId() == null
+                || user.getLogin() == null || user.getAge() == null
+                || user.getPassword() == null || user.getAge() < 18
+                || user.getAge() <= 0 || user.getLogin().length() < 6
+                || user.getPassword().length() < 6 || user.getLogin().contains(user.getLogin())
+                || !user.getPassword().contains("-?\\d+(\\.\\d+)?")
+                || Character.isLetter(user.getLogin().charAt(0))) {
+            throw new RuntimeException("Something went wrong...");
+        }
+
+        return user;
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -6,6 +6,16 @@
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null || user.getId() == null
+                || user.getLogin() == null || user.getAge() == null
+                || user.getPassword() == null || user.getAge() < 18
+                || user.getAge() <= 0 || user.getLogin().length() < 6
+                || user.getPassword().length() < 6 || user.getLogin().contains(user.getLogin())
```
MaliukDaria, line 13: `user.getLogin().contains(user.getLogin())` this check does not make sense, we need to check if there is a user with this login in the storage. We need to use method `User get (String login)` of StorageDaoImpl class for this purpose

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -6,6 +6,16 @@
...
+                || !user.getPassword().contains("-?\\d+(\\.\\d+)?")
+                || Character.isLetter(user.getLogin().charAt(0))) {
```
MaliukDaria, line 15: Why do we need this check?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -6,6 +6,16 @@
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null || user.getId() == null
+                || user.getLogin() == null || user.getAge() == null
+                || user.getPassword() == null || user.getAge() < 18
+                || user.getAge() <= 0 || user.getLogin().length() < 6
```
MaliukDaria, line 12: do we need this check (user.getAge() <= 0) if we have this one user.getAge() < 18 ?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,106 @@
...
+    void registerIdNull_notOK() {
+        assertThrows(RuntimeException.class, () -> {
+            registrationService.register(new User(null, "john34", "password", 18));
+        });
+    }
+
+    @Test
+    void registerLoginNull_notOk() {
+        assertThrows(RuntimeException.class, () -> {
+            registrationService.register(new User(1345L, null, "password", 18));
```
MaliukDaria, line 29: we need to register user without id

MaliukDaria (changes requested):

PR #102, opened 2022-02-08
Tests implemented. Coverage 93%. Codestyle passed.

lugburz-hr:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,11 +1,49 @@
...
+        return storageDao.get(user.getLogin()) == null ? storageDao.add(user) : null;
+    }
+
+    public void userExists(User user) {
+        if (user == null) {
+            throw new RuntimeException("There is no user");
+        }
+    }
+
+    public void validateLogin(User user) {
```
black-bit0, line 31: better to use boolean instead of void, they are easier to test

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private static RegistrationServiceImpl registrationService;
+    private static User user;
+
+    @BeforeAll
+    static void beforeAll() {
+        registrationService = new RegistrationServiceImpl();
+        user = new User();
+    }
+
```
black-bit0, line 23: to avoid setting values ​​for the user each time, you can set valid values ​​in beforeEach

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,138 @@
...
+    @Test
+    void register_isAgeCorrect_NotOk() {
+        user.setAge(-5);
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
+        user.setAge(102);
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
+        user.setAge(null);
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
+    }
+
```
black-bit0, line 38: naming error, after "_" write the first word in lower case

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,138 @@
...
+    @BeforeEach
+    void setUp() {
+        Storage.people.clear();
+    }
+
+    @Test
+    void register_isAgeCorrect_NotOk() {
+        user.setAge(-5);
```
black-bit0, line 31: it is good practice to point to CONSTANT_AGE / PASSWRORD_LENGTH - 1

lugburz-hr: -5 is random age < 0, could be -10 or -999 etc.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
lugburz-hr: these methods will throw RuntimeException, otherwise program flow will continue, there is no sense for them to return any value. if there would be some checkable exception, it makes sense, we must know that we got problem here, but it's RuntimeException, program will be terminated.

black-bit0 (approved):

black-bit0 (approved):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,11 +1,49 @@
...
+    private final StorageDaoImpl storageDao;
+
+    public RegistrationServiceImpl() {
+        storageDao = new StorageDaoImpl();
+    }
 
     @Override
     public User register(User user) {
-        return null;
+        userExists(user);
```
KostyaNikiforov, line 18:
```suggestion
        checkIfUserExists(user);
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,128 @@
...
+        user.setLogin("Third user");
+        assertEquals(user, registrationService.register(user));
+    }
+
+    @Test
+    void register_passwordIsIncorrect_notOk() {
+        user.setPassword("12345");
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
+        user.setPassword("123");
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
```
KostyaNikiforov, line 62: It is redundant because you already check if password is shorter than 6 letters

KostyaNikiforov (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,125 @@
...
+        fourthUser.setAge(100);
+        fourthUser.setPassword("vn3e89r38*");
+        fourthUser.setLogin("Fourth user");
+        assertEquals(fourthUser, registrationService.register(fourthUser));
+    }
+
+    @Test
+    void register_userIsNull_exception() {
+        user = null;
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
```
KostyaNikiforov, line 115:
```suggestion
        assertThrows(RuntimeException.class, () -> registrationService.register(null));
```

KostyaNikiforov (approved):

PR #126, opened 2022-04-16
- was implemented method register(User user)

MarkShapka:
- created unit tests

### DIFF
--- a/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
+++ b/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
@@ -1,0 +1,81 @@
+package core.basesyntax.service;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import core.basesyntax.dao.StorageDao;
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private User user1 = new User();
+    private User user2 = new User();
+    private User user3 = new User();
+    private StorageDao storageDao = new StorageDaoImpl();
+    private RegistrationService registrationService = new RegistrationServiceImpl();
+
+    @BeforeEach
+    void setUp() {
+        user1.setLogin("userLogin");
+        user1.setPassword("password");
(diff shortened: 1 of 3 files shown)


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,81 @@
...
+        user2.setLogin("userLogin2");
+        user2.setPassword("password2");
+        user2.setAge(16);
+        user3.setLogin("userLogin3");
+        user3.setPassword("pass");
+        user3.setAge(20);
+    }
+
+    @Test
+    void userRegistered() {
```
MaliukDaria, line 33: what about method naming? (There are a lot of ways to name your test methods. The main point is that they should have informative names and be consistent along with other developers in your team. For this task use such convention: <methodUnderTest>_<state>_<expectedBehavior>; For example, if we are testing the method register with a null)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,81 @@
...
+        } catch (RuntimeException e) {
+            return;
+        }
+        fail("test should fail -> user is null");
+    }
+
+    @Test
+    void checkUserLoginWithSameLogin() {
+        storageDao.add(user1);
+        storageDao.add(user2);
```
MaliukDaria, line 63: I think it's better to use `register` method here

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,81 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import core.basesyntax.dao.StorageDao;
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
```
MaliukDaria, line 12: I think we could add more tests. For example checks for:
- null age, name, password
- empty name, password
- 0 or negative age 
 and so on..

MaliukDaria (changes requested):

MaliukDaria (changes requested):
test

MaliukDaria (approved):

PR #195, opened 2022-06-14
tests were written

mykola8426:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,26 @@
...
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user.getAge() < 18) {
+            throw new RuntimeException("Age must be more than 17");
+        }
+        if (user.getPassword() == null || user.getPassword().length() < 6) {
+            throw new RuntimeException("Password must be more than 5 symbols");
```
NikitaSemisoshenko, line 17: All magic number should be constants. Lets create **private static final int MINIMUM_PASSWORD_LENGTH = 6** field and use it instead of "6" in this case.

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,97 @@
...
+    @BeforeEach
+    void setUp() {
+        firstUser = new User("Mike", "mike12345", 22);
+        secondUser = new User("Alice", "password", 19);
+        thirdUser = new User("Melisa", "qwerty", 20);
+        fourthUser = new User("MaX", "MaX24MaX", 24);
+    }
+
+    @Test
+    void password_Length_Is_Valid() {
```
NikitaSemisoshenko, line 39: all parts of method title sholud starts with small letter. Also in the end should be writed result of this test. Lets change for **"password_lengthIsValid_Ok"**. This comment is valid for this and other test methods.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,26 @@
...
+import core.basesyntax.db.Storage;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user.getAge() < 18) {
```
NikitaSemisoshenko, line 13: All magic numbers should be constants.

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,97 @@
...
+        firstUser.setPassword("ddd");
+        secondUser.setPassword("asdadado");
+        thirdUser.setPassword(null);
```
NikitaSemisoshenko, line 42: Let's create 3 different test methods for 3 differrent cases: short password, valid password and null password.

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,97 @@
...
+        assertEquals(secondUser, registrationService.register(secondUser));
+        assertThrows(RuntimeException.class, () -> registrationService.register(thirdUser));
+    }
+
+    @Test
+    void age_Is_Valid() {
+        firstUser.setAge(10);
+        secondUser.setAge(22);
+        thirdUser.setAge(null);
+        fourthUser.setAge(-100);
```
NikitaSemisoshenko, line 53: Same case as with passwords: create one test method for one test case. Don't combine them all at one method. It's actual thing for all of your test cases.

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Test
+    void check_Addition_To_List() {
+        fourthUser.setLogin("Mike");
+
+        registrationService.register(firstUser);
+        assertEquals(1, Storage.people.size());
+        registrationService.register(secondUser);
+        assertEquals(2, Storage.people.size());
+        registrationService.register(thirdUser);
```
NikitaSemisoshenko, line 69: In case when you did register one user and test method shows correct work of your solution you don't need repeat and check register operation second time. One case will be enough.

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,97 @@
...
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.User;
+import core.basesyntax.service.RegistrationService;
+import core.basesyntax.service.RegistrationServiceImpl;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
```
NikitaSemisoshenko, line 18: let's make this class and all of its methods public

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
NikitaSemisoshenko, line 13: Also what about if we separate register and check user values methods for better code readability?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,26 @@
...
+        if (user.getPassword() == null || user.getPassword().length() < 6) {
+            throw new RuntimeException("Password must be more than 5 symbols");
+        }
+        for (User userFromList : Storage.people) {
+            if (userFromList.getLogin().equals(user.getLogin())) {
+                throw new RuntimeException("User login already exist");
+            }
+        }
+        storageDao.add(user);
+        return user;
```
olyaklipa, line 25:
```suggestion
        return storageDao.add(user);
```

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
olyaklipa, line 39: test method names should follow the next rule: methodUnderTest_state_expectedBehavior
So, for example, I would call this method register_passwordLength_isValid

olyaklipa (changes requested):

olyaklipa (approved):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -11,16 +11,28 @@ public class RegistrationServiceImpl implements RegistrationService {
 
     @Override
     public User register(User user) {
+        checkAge(user);
+        passwordCheck(user);
+        loginExistsCheck(user);
+        storageDao.add(user);
```
NikitaSemisoshenko: What about if user will input null login?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
KostyaNikiforov: Agree with Nikita, you do overmuch work in almost every test method

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+    @BeforeEach
+    public void setUp() {
+        firstUser = new User("Mike", "mike12345", 22);
+        secondUser = new User("Alice", "password", 19);
+        thirdUser = new User("Melisa", "qwerty", 20);
+        fourthUser = new User("MaX", "MaX24MaX", 24);
+    }
+
+    @Test
+    public void register_passwordLength_isValid() {
```
KostyaNikiforov, line 39: Fix test mehod names according to readme
```suggestion
    public void register_passwordLength_ok() {

```

NikitaSemisoshenko (approved):

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+        firstUser.setAge(10);
+        secondUser.setAge(null);
+        thirdUser.setAge(17);
+        fourthUser.setAge(-100);
+
+        assertThrows(RuntimeException.class, () -> registrationService.register(firstUser));
+        assertThrows(RuntimeException.class, () -> registrationService.register(secondUser));
+        assertThrows(RuntimeException.class, () -> registrationService.register(thirdUser));
+        assertThrows(RuntimeException.class, () -> registrationService.register(fourthUser));
+    }
```
KostyaNikiforov, line 73: The first part of the test method name should be the name of the method that you test;
The second part should describe the state you pass to this method;
And the last part should describe the expected behavior;
For example: 
register_nullPasword_notOk
register_existingUser_notOk

Let's divide your test methods into several smaller ones with proper names

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+        StorageDao getUserBylogin = new StorageDaoImpl();
+        assertEquals(firstUser, getUserBylogin.get("Mike"));
+        assertEquals(secondUser, getUserBylogin.get("Alice"));
+        assertEquals(thirdUser, getUserBylogin.get("Melisa"));
+        assertEquals(fourthUser, getUserBylogin.get("MaX"));
+    }
+
+    @Test
+    public void getLoginCheck_isNotValid() {
```
KostyaNikiforov, line 108: Fix naming

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @BeforeEach
+    public void setUp() {
+        firstUser = new User("Mike", "mike12345", 22);
+        secondUser = new User("Alice", "password", 19);
+        thirdUser = new User("Melisa", "qwerty", 20);
+        fourthUser = new User("MaX", "MaX24MaX", 24);
+    }
```
KostyaNikiforov, line 36: You can create only one user and just change its fields for each test case

KostyaNikiforov (changes requested):

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,98 @@
...
+        StorageDao getUserBylogin = new StorageDaoImpl();
+        assertEquals(user, getUserBylogin.get("Mike"));
+    }
+
+    @Test
+    public void get_byLogin_notOk() {
+        StorageDao getUserBylogin = new StorageDaoImpl();
+        assertNull(getUserBylogin.get("AAAA"));
+        assertNull(getUserBylogin.get(null));
+    }
```
KostyaNikiforov, line 86: we do not need to test dao classes

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,98 @@
...
+    public void addToList_length_ok() {
+        registrationService.register(user);
+        assertEquals(1, Storage.people.size());
+    }
+
+    @Test
+    public void get_byLogin_ok() {
+        registrationService.register(user);
+
+        StorageDao getUserBylogin = new StorageDaoImpl();
```
KostyaNikiforov, line 77: you can get an access to the storage diractly

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Test
+    public void getLoginCheck_isValid() {
+        registrationService.register(firstUser);
+        registrationService.register(secondUser);
+        registrationService.register(thirdUser);
+        registrationService.register(fourthUser);
+
+        StorageDao getUserBylogin = new StorageDaoImpl();
```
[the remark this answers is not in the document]
KostyaNikiforov: +

KostyaNikiforov (changes requested):

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Test
+    public void register_nullPassword_0k() {
+        user.setPassword(null);
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    public void register_age_notOk() {
+        user.setAge(10);
```
KostyaNikiforov, line 49: Let's take the maximum invalid value.
In this case it's 17

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Test
+    public void register_passwordLength_ok() {
+        user.setPassword("dddssssssssad");
+        assertEquals(user, registrationService.register(user));
+    }
+
+    @Test
+    public void register_passwordLength_notOk() {
+        user.setPassword("ddd");
```
KostyaNikiforov, line 37: The same here, password length should be maxing invalid

KostyaNikiforov (changes requested):
It would be also cool to add a test case when the user is already registered

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Test
+    public void register_nullAge_ok() {
+        user.setAge(null);
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    public void addToList_length_ok() {
```
KostyaNikiforov, line 65: Please, fix naming and add the test which I suggested before

KostyaNikiforov (changes requested):

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Test
+    public void register_nullAge_ok() {
+        user.setAge(null);
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    public void size_addElement_biggerSize() {
```
KostyaNikiforov, line 65:
```suggestion
    public void register_validUser_ok() {

```

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+    @Test
+    public void register_age_notOk() {
+        user.setAge(17);
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    public void register_age_ok() {
+        assertEquals(user, registrationService.register(user));
+    }
```
KostyaNikiforov, line 56: You can combine that method with this one, because they do the same things

KostyaNikiforov (changes requested):
Add a test method that will check if there is a user with the same email in a storage

KostyaNikiforov (changes requested):
What about this message?
Screenshot 2022-06-15 at 11 06 06

KostyaNikiforov (approved):
Good job!)🔥

PR #199, opened 2022-06-16
tests and impl were created

Evhenii-Vaskiv:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -9,6 +9,15 @@ public class RegistrationServiceImpl implements RegistrationService {
...
     @Override
     public User register(User user) {
-        return null;
+        if (user == null || user.getLogin() == null
+                || user.getPassword() == null || user.getAge() == null) {
+            throw new RuntimeException("There is invalid user`s data.");
+        }
+        if (user.getAge() < 18 || (storageDao.get(user.getLogin()) != null)
+                || (user.getPassword().length() < 6)) {
+            return null;
```
MaliukDaria, line 18: why `return null`? I think it's better to throw some exception here

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -9,6 +9,15 @@ public class RegistrationServiceImpl implements RegistrationService {
...
+        }
+        storageDao.add(user);
+        return user;
```
MaliukDaria, line 21:
```suggestion
        return storageDao.add(user);
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @BeforeEach
+    void setUp() {
+        john.setAge(19);
+        john.setLogin("John12.03.1993");
+        john.setPassword("12345678");
+    }
+
+    @Test
+    void validUser_Ok() {
```
MaliukDaria, line 31: There are a lot of ways to name your test methods. The main point is that they should have informative names and be consistent along with other developers in your team. For this task use such convention: `<methodUnderTest>_<state>_<expectedBehavior>`; For example, if we are testing the method register with a `null` user's age the test method name should be `register_nullAge_notOk`. `notOk` is because the test expects the register method to throw an exception.
```suggestion
    void register_validUser_Ok() {
```

same for other methods

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+        registrationService.register(john);
```
MaliukDaria, line 32: lets also check is user that returned from register method is the same as we want to register(have the same fields)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+        vanya.setLogin("John12.03.1993");
+        vanya.setPassword("dfsndlfwe19383");
+        int expected = Storage.people.size();
+        registrationService.register(vanya);
+        int actual = Storage.people.size();
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    void lessThanSixSymbolsPassword_Ok() {
```
MaliukDaria, line 61: same, we should expect some exception here

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+            registrationService.register(john);
+        });
+        john = new User();
+    }
+
+    @AfterEach
+    void tearDown() {
+        for (int i = 0; i < Storage.people.size(); i++) {
+            Storage.people.remove(i);
+        }
```
MaliukDaria, line 106: Lets use `clear()` method here

MaliukDaria (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,110 @@
...
+    @Test
+    void register_userWithLessAge_NotOk() {
+        john.setAge(17);
+        assertThrows(ValidationException.class, () -> {
+            registrationService.register(john);
+        });
+    }
+
+    @Test
+    void register_UserWithSameLogin_NotOk() {
```
Nick97-git, line 54:
```suggestion
    void register_userWithSameLogin_NotOk() {
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,110 @@
...
+    void register_nullAgeUser_NotOk() {
+        john.setAge(null);
+        assertThrows(RuntimeException.class, () -> {
+            registrationService.register(john);
+        });
+    }
+
+    @Test
+    void register_nullUser_NotOk() {
+        john = null;
```
Nick97-git, line 99: don't use john here and pass directly null

Nick97-git (changes requested):

PR #204, opened 2022-06-26
add solution

uglovyi:


src/main/java/core/basesyntax/model/User.java
```diff
@@ -3,6 +3,7 @@
 import java.util.Objects;
 
 public class User {
+    public static final Integer ADULT_AGE = 18;
```
AlexRozv, line 6: instead of adding ADULT_AGE to the User class let's just create constants in the classes where we use this number

AlexRozv (changes requested):

PR #206, opened 2022-06-27
JUnit test and implementation method are created according to readme …

antkevyv:
…file.

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -1,11 +1,26 @@
 package core.basesyntax.service;
 
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.db.Storage;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private StorageDaoImpl storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user.getPassword().length() < 6) {
+            throw new RuntimeException("Password length should be 6 or more characters");
+        }
+        if (user.getAge() < 18) {
+            throw new RuntimeException("User should be at least 18 years old");
+        }
+        for (User user1 : Storage.people) {
+            if (user1.equals(user)) {
+                throw new RuntimeException("User already exists");
+            }
+        }
+        storageDao.add(user);
+        return user;
     }
 }
--- a/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
+++ b/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
@@ -1,0 +1,63 @@
+package core.basesyntax.service;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private final RegistrationService registrationService = new RegistrationServiceImpl();
+    private final User user = new User();
+
+    @BeforeEach
+    void setUp() {
+        user.setLogin("admin");
+        user.setAge(18);
+        user.setPassword("123456");
+        user.setId(13L);
+    }
+
+    @Test
+    void correctUserAge_Ok() {
+        Integer actual = user.getAge();
+        assertFalse(actual < 18);
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,11 +1,27 @@
...
+import core.basesyntax.dao.StorageDao;
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.db.Storage;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
```
MaliukDaria, line 12: lets also add check for user == null, and for users fields ane null

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,11 +1,27 @@
...
-        return null;
+        if (user.getPassword().length() < 6) {
+            throw new RuntimeException("Password length should be 6 or more characters");
+        }
+        if (user.getAge() < 18) {
+            throw new RuntimeException("User should be at least 18 years old");
+        }
```
MaliukDaria, line 18: lets make numbers 6 and 18 as constants

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,63 @@
...
+    @BeforeEach
+    void setUp() {
+        user.setLogin("admin");
+        user.setAge(18);
+        user.setPassword("123456");
+        user.setId(13L);
+    }
+
+    @Test
+    void correctUserAge_Ok() {
```
MaliukDaria, line 24: There are a lot of ways to name your test methods. The main point is that they should have informative names and be consistent along with other developers in your team. For this task use such convention: `<methodUnderTest>_<state>_<expectedBehavior>`; For example, if we are testing the method register with a `null` user's age the test method name should be `register_nullAge_notOk`. `notOk` is because the test expects the register method to throw an exception.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,63 @@
...
+        Integer actual = user.getAge();
+        assertFalse(actual < 18);
```
MaliukDaria, line 26: what does exactly we test here?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Test
+    void correctUserAge_Nok() {
+        User user1 = new User();
+        user1.setAge(17);
```
MaliukDaria, line 32: If we want to test users age, we should set all valid else fields except age

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,63 @@
...
+        assertThrows(RuntimeException.class, () -> {
+            boolean b = registrationService.register(user1).getAge() < 18;
+        });
+    }
+
+    @Test
+    void correctUserPasswordLength_Ok() {
+        int actual = user.getPassword().length();
+        assertTrue(actual >= 6);
+    }
```
MaliukDaria, line 43: same, what do we test here?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,63 @@
...
+            boolean b = registrationService.register(user1).getPassword().length() < 6;
+        });
+    }
+
+    @Test
+    void repeatedUser_Nok() {
+        User user1 = new User();
+        user1.setLogin("admin");
+
+        assertThrows(RuntimeException.class, () -> registrationService.register(user1)
```
MaliukDaria, line 60: why we should get exception here?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
```
MaliukDaria, line 11: We also need to clear storage after every test

MaliukDaria (changes requested):

PR #233, opened 2022-08-09
implemented RegistrationServiceImpl, created RegistrationServiceImplT…

NataliaKurhanska:
…est class


zachar-chorny (changes requested):
Also in all unsuccessful tests, check if the Storage has changed, because just because we get an error does not guarantee that the method did not register the user.

aaananas (approved):

PR #247, opened 2022-08-18
added solution

Nesterenko-Andrey:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,31 @@
...
+    private static final int MIN_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new RuntimeException("User can`t be null");
+        }
+        if (user.getLogin() == null || user.getLogin().isEmpty()) {
```
justpopova, line 17:
```suggestion
        if (user.getLogin() == null || user.getLogin().isEmpty()) {
```

maybe you will try to add here checking, if login contains/matches empty spaces? because its not good if someone will make his/her login "S    h    i ".(between each letter at least 2-3 spaces, but here can't see, don't know why)

justpopova (changes requested):
Goood job!

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+    @BeforeEach
+    void setUp() {
+        user = new User();
+        user.setLogin("ALice");
+        user.setPassword("456lfkdgj");
+        user.setAge(22);
+    }
+
+    @Test
+    void register_emptyUser_NotOk() {
```
justpopova, line 35:
```suggestion
    void emptyUser_NotOk() {
```
also maybe don't add register everywhere, already understandable that this test is for registration method, but its up to you^_^

justpopova (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
Nesterenko-Andrey: I think that login can be "Anna Popova" and in this case empty spaces can be needed

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Nesterenko-Andrey: fixed)

justpopova (changes requested):
🔥

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
justpopova: Okay, cause i thought login needs to be like login in IG or twitter ahaha no spaces can add:)))

justpopova (approved):
good!

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+        user2.setPassword("slkgnUR12");
+        user2.setAge(20);
+        Assertions.assertThrows(RuntimeException.class, () -> registrationService.register(user2));
+    }
+
+    @Test
+    void lowAge_NotOk() {
+        user.setAge(15);
+        Assertions.assertThrows(RuntimeException.class, () -> registrationService.register(user));
+    }
```
dSkarzh, line 66:
```suggestion
    @Test
    void register_lessThanMinAge_NotOk() {
        user.setAge(MIN_AGE - 1);
        Assertions.assertThrows(RuntimeException.class, () -> registrationService.register(user));
    }
```
add method name to test case names
use closest data to the requirements (if age limit is 18 then edge data is 17 18 19 -1)
create constant for MIN_AGE, it can be used several times and will improve readibility

dSkarzh: add test for age larger than min

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+    void minAcceptableAge_Ok() {
+        user.setAge(18);
+        assertEquals(user, registrationService.register(user));
+    }
+
+    @Test
+    void shortPassword_NotOk() {
+        user.setPassword("pass");
+        Assertions.assertThrows(RuntimeException.class, () -> registrationService.register(user));
+    }
```
dSkarzh, line 90: add test for min required length password, and more the min required length

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+    @Test
+    void existingUser_NotOk() {
+        storageDao.get(user.getLogin());
+        storageDao.add(user);
+        User user2 = new User();
+        user2.setLogin(user.getLogin());
+        user2.setPassword("slkgnUR12");
+        user2.setAge(20);
+        Assertions.assertThrows(RuntimeException.class, () -> registrationService.register(user2));
+    }
```
dSkarzh, line 60: create two users: firstUser and secondUser (using numbers in variable names is a bad practice)
register them by calling `register()`, not `storageDao.add()`
the second time you call `register()` you should expect error to be thrown

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static RegistrationService registrationService;
+    private static StorageDao storageDao;
```
dSkarzh, line 17: you shouldn't use dao, because breaks isolation principle of unit tests, use storage directly if you need to

dSkarzh (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,122 @@
...
+    void register_moreThanMinAcceptableAge_Ok() {
+        user.setAge(MIN_AGE + 1);
+        assertEquals(user, registrationService.register(user));
+    }
+
+    @Test
+    void register_shortPassword_NotOk() {
+        user.setPassword("pass");
+        Assertions.assertThrows(RuntimeException.class, () -> registrationService.register(user));
+    }
```
Noct2000, line 93: it would be better to use a wrong password that more closer to valid value

Noct2000 (changes requested):

Noct2000 (approved):

PR #250, opened 2022-08-19
Dev

max-bolotin:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,49 @@
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.db.Storage;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_AGE = 18;
+    private static final int MAX_AGE = 99;
```
AshaJaneBrown, line 10: there is not such requirement, but let it be)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,49 @@
...
+            throw new NullPointerException("Login should not be null");
+        }
+        if (user.getPassword() == null) {
+            throw new NullPointerException("Password should not be null");
+        }
+        for (User userLogin : Storage.people) {
+            if (userLogin.getLogin().equals(user.getLogin())) {
+                throw new RuntimeException("User should have unique login");
+            }
+        }
```
AshaJaneBrown, line 32: we can use storageDao get() method for that purpose as well

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static RegistrationService registrationService;
+    private static StorageDao storageDao;
```
AshaJaneBrown, line 17: due to mentors comments, we should not use storageDao in this class

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+        storageDao = new StorageDaoImpl();
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.people.clear();
+    }
+
+    @Test
+    public void registerAndGetNewUser_Ok() {
```
AshaJaneBrown, line 32: I suggest test name to be like that: register_validData_Ok, the firsrt part is action - register, second part describes data - validData

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+        User userFirst = createUser(22, "crimeabridgedestroyed", "Himars");
+        User userSecond = createUser(70, "freedom", "Bayraktar");
```
AshaJaneBrown, line 34: why do we register a second user here? one should be enough

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+        registrationService.register(userFirst);
+        User expectedUser1 = storageDao.get(Storage.people.get(0).getLogin());
+        assertEquals(userFirst, expectedUser1);
+        registrationService.register(userSecond);
+        User expectedUser2 = storageDao.get(Storage.people.get(1).getLogin());
+        assertEquals(userSecond, expectedUser2);
+    }
+
+    @Test
+    void registerAndGetDifferentUsers_Ok() {
```
AshaJaneBrown, line 44: due to mentors comment, there is no need to create test case for 2 different users

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+        User userSecond = createUser(70, "freedom", "Bayraktar");
+        registrationService.register(userFirst);
+        registrationService.register(userSecond);
+        User expectedUser1 = storageDao.get(Storage.people.get(0).getLogin());
+        User expectedUser2 = storageDao.get(Storage.people.get(1).getLogin());
+        assertNotEquals(expectedUser1, expectedUser2);
+    }
+
+    @Test
+    void userNull_GetExceptionMessage_Ok() {
```
AshaJaneBrown, line 55:
```suggestion
    void register_nullUser_NotOk() {
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,49 @@
...
+    private static final int MIN_AGE = 18;
+    private static final int MAX_AGE = 99;
+    private static final int MIN_PSSWD_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new NullPointerException("User should not be null");
```
AshaJaneBrown, line 17: i am not sure, but shouldn't we use RuntimeException instead?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+            registrationService.register(userNull);
+        } catch (NullPointerException e) {
+            String actualMessage = e.getMessage();
+            String expectedMessage = "User should not be null";
+            assertEquals(actualMessage, expectedMessage);
+        }
+    }
+
+    @Test
+    void userOneDouble_ExceptionMessage_Ok() {
```
AshaJaneBrown, line 66: unclear name

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+        User userFirst = createUser(22, "crimeabridgedestroyed", "Himars");
+        User userSecond = createUser(70, "freedom", "Bayraktar");
+        registrationService.register(userFirst);
+        registrationService.register(userSecond);
+        User userDoubleFirst = createUser(22, "crimeabridgedestroyed", "Himars");
```
AshaJaneBrown, line 71: repeated data for valid user can be constants, like that
    public static final String DEFAULT_LOGIN = "login";
    public static final String DEFAULT_PASSWORD = "password";
    public static final int DEFAULT_AGE = 30;

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+    @Test
+    void userNullLogin_Exception_Ok() {
+        User userNullLogin = createUser(80, "crimeabridgedestroyed", null);
+        assertThrows(NullPointerException.class, () -> registrationService.register(userNullLogin));
+    }
+
+    @Test
+    void userNullLogin_ExceptionMessage_Ok() {
+        User userNullLogin = createUser(80, "crimeabridgedestroyed", null);
```
AshaJaneBrown, line 190: we can place valid user creation in the setUp block, for example
   @BeforeEach
    void setUp() {
        validUser = new User();
        validUser.setLogin(DEFAULT_LOGIN);
        validUser.setPassword(DEFAULT_PASSWORD);
        validUser.setAge(DEFAULT_AGE);
    }

AshaJaneBrown (changes requested):
good job!

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
max-bolotin: I just want to check if I can add more than one user and then get them all with no mistakes

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
max-bolotin: Why is it so? There might be a situation when the second user overwrites the first one, for example. I want to test all the possible mistakes

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
max-bolotin: I'm not sure either. We are checking the null value, so I prefer to keep NPE unless mentor tells me to change it.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
max-bolotin: I don't think I should do it because I don't need a valid user before each test. I have many tests that check invalid fields, and creating a valid user each time will waste resources

AshaJaneBrown (approved):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,43 @@
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_AGE = 18;
+    private static final int MAX_AGE = 99;
```
Noct2000, line 9: this is redundant check

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,43 @@
...
+        if (user.getLogin() == null) {
+            throw new NullPointerException("Login should not be null");
+        }
+        if (user.getPassword() == null) {
+            throw new NullPointerException("Password should not be null");
+        }
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new RuntimeException("User should have unique login");
+        }
+        if (user.getAge() < MIN_AGE && user.getAge() >= 0) {
```
Noct2000, line 30:
```suggestion
        if (user.getAge() < MIN_AGE) {
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,43 @@
...
+            throw new RuntimeException("User too young!");
+        }
+        if (user.getAge() > MAX_AGE) {
+            throw new RuntimeException("User too old!");
+        }
+        if (user.getAge() < 0) {
+            throw new RuntimeException("User's age cannot be negative!");
```
Noct2000, line 37: I think this is redundant we already have `throw new RuntimeException("User too young!");`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,43 @@
...
+        }
+        if (user.getPassword().length() < MIN_PSSWD_LENGTH) {
+            throw new RuntimeException("Password too short!");
+        }
+        if (user.getLogin().length() == 0) {
+            throw new RuntimeException("Login should not be blank!");
```
Noct2000, line 43: it would be better to use `.isBlank()`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static RegistrationService registrationService;
+    private static final User userNull = null;
```
Noct2000, line 15: I think this is redundant

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,254 @@
...
+        assertEquals(3, actual);
+    }
+
+    private User createUser(Integer age, String password, String login) {
+        User user = new User();
+        user.setAge(age);
+        user.setPassword(password);
+        user.setLogin(login);
+        return user;
+    }
```
Noct2000, line 252: it would be better to create new constructor for `User`

Noct2000 (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,213 @@
...
+        User userTooYoung = createUser(10, DEFAULT_PASSWORD_FIRST, DEFAULT_LOGIN_FIRST);
+        try {
+            registrationService.register(userTooYoung);
+        } catch (RuntimeException e) {
+            String actualMessage = e.getMessage();
+            String expectedMessage = "User too young!";
+            assertEquals(actualMessage, expectedMessage);
+        }
+    }
+
```
Noct2000, line 105: one from these methods is redundant, you don't need check one case several times and use values closer to valid values, check another methods

Noct2000 (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,157 @@
...
+        registrationService.register(userGoodFirst);
+        registrationService.register(userGoodSecond);
+        registrationService.register(userGoodThird);
+        int actual = Storage.people.size();
+        assertEquals(3, actual);
+    }
+
+    private User createUser(Integer age, String password, String login) {
+        return new User(login, password, age);
+    }
```
Noct2000, line 155: use `new User(login, password, age);` directly

Noct2000 (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,152 @@
...
+        storageDao = new StorageDaoImpl();
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.people.clear();
+    }
+
+    @Test
+    public void register_GetNewUser_Ok() {
```
Noct2000, line 33:
```suggestion
    public void register_getNewUser_Ok() {
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,152 @@
...
+        } catch (RuntimeException e) {
+            String actualMessage = e.getMessage();
+            String expectedMessage = "User should have unique login";
+            assertEquals(actualMessage, expectedMessage);
+        }
+    }
+
+    @Test
+    void register_userTooYoungExceptionMessage_Ok() {
+        User userTooYoung = new User(10, DEFAULT_PASSWORD_FIRST, DEFAULT_LOGIN_FIRST);
```
Noct2000, line 71:
```suggestion
        User userTooYoung = new User(17, DEFAULT_PASSWORD_FIRST, DEFAULT_LOGIN_FIRST);
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,152 @@
...
+            registrationService.register(userEmptyLogin);
+        } catch (RuntimeException e) {
+            String actualMessage = e.getMessage();
+            String expectedMessage = "Login should not be blank!";
+            assertEquals(actualMessage, expectedMessage);
+        }
+    }
+
+    @Test
+    void register_GetBaseSize_Ok() {
```
Noct2000, line 142:
```suggestion
    void register_getBaseSize_Ok() {
```

Noct2000 (changes requested):

Noct2000 (approved):

PR #251, opened 2022-08-19
add solution

AndrewHudzenko:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,109 @@
...
+    void register_minAge_Ok() {
+        user.setAge(MIN_AGE);
+        assertEquals(user, registrationService.register(user));
+    }
+
+    @Test
+    void register_moreThenMinAge_Ok() {
+        user.setAge(MIN_AGE);
+        assertEquals(user, registrationService.register(user));
+    }
```
max-bolotin, line 50: What's the difference between these two methods: register_minAge_Ok() and register_moreThenMinAge_Ok() ?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Test
+    void register_nullAge_NotOk() {
+        user.setAge(0);
+        checkException(user);
+    }
```
max-bolotin, line 56: Null and zero aren't the same. Either the method naming or the age value are incorrect

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,109 @@
...
+    @AfterEach
+    void tearDown() {
+        Storage.people.clear();
+    }
+
+    private void checkException(User currentUser) {
+        assertThrows(ValidationException.class,
+                () -> registrationService.register(user));
+    }
+}
```
max-bolotin, line 109: You might want to add some null values check, for example null login or password

max-bolotin, line 109: You may add different users to the list and check their get method, as well as checking if they are really different

max-bolotin, line 109: Add the duplicate users adding check

max-bolotin, line 109: Add a check for the null user adding

max-bolotin, line 109: You may also add a check if you receive a corect exception message. This may be done somehow in this way:
```
try {
            registrationService.register(userDouble);
        } catch (ValidationException e) {
            String actualMessage = e.getMessage();
            String expectedMessage = "User should have unique login";
            assertEquals(actualMessage, expectedMessage);
        }
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,34 @@
...
+        return user;
+    }
+
+    private boolean checkUser(User user) {
+        if ((storageDao.get(user.getLogin()) == null)
+                && (user.getAge() >= MIN_AGE)
+                && (user.getPassword().length() >= MIN_PASSWORD_LENGTH)) {
+            return true;
+        }
+        return false;
```
max-bolotin, line 33: You might (or might not - I don't know if it is required) create different checks for different fields with different error messages, e.g. null login or password. I believe it's optional though, so it's up to you wheather to implement it or not

max-bolotin (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
AndrewHudzenko: All this ckecks already implemented in registrationService.register()

AndrewHudzenko: I don't fully understand why we need it with try catch if we already have assertThrows()

max-bolotin: assertThrows() only shows that the exception is thrown, not the exception message. Probably you really don't need it because you're using your custom exception. I used NPE and Runtime Exceptions so I wanted to be sure that the exception is thrown in a proper place and time in a code and that I receive a correct message. So it's up to you, I get your point.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
AndrewHudzenko: I added a general message if a field does not comply with the requirements, but if I add a message for each check, it would have to change the entire structure of the entire class, because in this case the boolean method will be useless and everything will need to be prescribed in the register method, because I can't add to each check error and return True or False

max-bolotin: I don't know if it is required, probably not, so for me it's ok if you don't want to re-write the entire structure of your class

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
max-bolotin: I mean tests for null values. I can see the test for null login, but no tests for null password or age

max-bolotin (approved):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,31 @@
...
+    }
+
+    private boolean checkUser(User user) {
+        if ((storageDao.get(user.getLogin()) == null)
+                && (user.getAge() >= MIN_AGE)
+                && (user.getPassword().length() >= MIN_PASSWORD_LENGTH)) {
+            return true;
+        } else {
+            throw new RuntimeException("Invalid data, check some fields");
+        }
```
Noct2000, line 33: it would be better to split user checks by criteria and move it to private methods and you don't need to check `id`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,133 @@
...
+        user = new User();
+        user.setAge(DEFAULT_AGE);
+        user.setLogin(DEFAULT_LOGIN);
+        user.setPassword(DEFAULT_PASSWORD);
+    }
+
+    @Test
+    void validUser_Ok() {
+        assertEquals(user, registrationService.register(user));
+    }
```
Noct2000, line 38: check naming of test methods `<methodUnderTest>_<state>_<expectedBehavior>`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,133 @@
...
+        assertNotEquals(expectedUser1, expectedUser2);
+    }
+
+    @Test
+    void register_userIsNull_NotOk() {
+        checkException(null);
+    }
+
+    @Test
+    void register_NullAge_NotOk() {
```
Noct2000, line 57: check case after `_`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,31 @@
...
+        return user;
+    }
+
+    private boolean checkUser(User user) {
+        if ((storageDao.get(user.getLogin()) == null)
+                && (user.getAge() >= MIN_AGE)
+                && (user.getPassword().length() >= MIN_PASSWORD_LENGTH)) {
+            return true;
+        } else {
+            throw new RuntimeException("Invalid data, check some fields");
```
Noct2000, line 32: we don't need else after return

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,31 @@
...
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (checkUser(user)) {
+            storageDao.add(user);
+        }
+        if (user.getLogin() == null
+                || user.getPassword() == null
```
Noct2000, line 18: check `null` first

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,133 @@
...
+        Storage.people.clear();
+    }
+
+    private User createUser(Integer age, String password, String login) {
+        User user = new User();
+        user.setAge(age);
+        user.setPassword(password);
+        user.setLogin(login);
+        return user;
+    }
```
Noct2000, line 127: you can create constructor for `User`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Test
+    void register_differentUsers_Ok() {
+        User user1 = createUser(DEFAULT_AGE, DEFAULT_PASSWORD, DEFAULT_LOGIN);
+        User user2 = createUser(DEFAULT_AGE, DEFAULT_PASSWORD, DEFAULT_LOGIN + "a");
+        registrationService.register(user);
+        registrationService.register(user2);
+        User expectedUser1 = Storage.people.get(0);
+        User expectedUser2 = Storage.people.get(1);
```
Noct2000, line 47: use `UserDao`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,133 @@
...
+        assertNotEquals(expectedUser1, expectedUser2);
```
Noct2000, line 48: use `assertEquals(expected, actual)` for each user

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,31 @@
...
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_PASSWORD_LENGTH = 6;
+    private static final int MIN_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (checkUser(user)) {
```
Noct2000, line 14: check `user == null`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Test
+    void register_emptyPassword_NotOk() {
+        user.setPassword("");
+        checkException(user);
+    }
+
+    @Test
+    void register_shortPassword_NotOk() {
+        user.setPassword("p");
```
Noct2000, line 100: it would be better to check password with 5 chars

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,133 @@
...
+    void register_minPasswordLength_Ok() {
+        user.setPassword("passwo");
+        assertEquals(user, registrationService.register(user));
+    }
+
+    @Test
+    void register_MoreThenMinPasswordLength_Ok() {
+        user.setPassword(DEFAULT_PASSWORD);
+        assertEquals(user, registrationService.register(user));
+    }
```
Noct2000, line 114: this tests are the same

Noct2000 (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
AndrewHudzenko: these are not the same tests

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
AndrewHudzenko: I don't know what it is and how to use it / I think it's something with JDBC / but I can't find any normal information about it and how to use it

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
AndrewHudzenko: ?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Noct2000: image

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Noct2000: your methods must have familiar names, if you use `register_minAge_Ok` than `register_NullAge_NotOk` must be `register_nullAge_NotOk`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,140 @@
...
+        user.setPassword("passw");
+        checkException(user);
+    }
+
+    @Test
+    void register_minPasswordLength_Ok() {
+        user.setPassword("passwo");
+        assertEquals(user, registrationService.register(user));
+    }
+
```
Noct2000, line 116: this is one wrong use case

Noct2000 (changes requested):
check old comments

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
AndrewHudzenko: If I understand correctly, in this case I have to change the first letter after _ to a lowercase letter?

Noct2000 (changes requested):
check old comments

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+        Storage.people.clear();
+    }
+
+    private User createUser(Integer age, String password, String login) {
+        User user = new User();
+        user.setAge(age);
+        user.setPassword(password);
+        user.setLogin(login);
+        return user;
+    }
```
Noct2000, line 131: replace it method with constructor `User(Integer age, String password, String login)` in the `User`

Noct2000 (changes requested):

Noct2000 (approved):

PR #262, opened 2022-08-21
added tests and method realisation

Salmutter:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,28 @@
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    public static final int MIN_AGE = 18;
+    public static final int MIN_PASSWORD_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
```
MaliukDaria, line 13: lets also add check for user == null

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,28 @@
...
-        return null;
+        if (user.getPassword() == null || user.getLogin() == null || user.getAge() == null) {
+            throw new NullPointerException("Value cant be null");
```
MaliukDaria, line 15:
```suggestion
            throw new RuntimeException("Value cant be null");
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Test
+    void register_emptyLogin_notOk() {
+        user.setLogin("");
+        assertThrows(RuntimeException.class, () -> registration.register(user),
+                "User login can`t be empty");
+    }
+
+    @Test
+    void register_existingLogin_notOk() {
+        registration.register(user);
```
MaliukDaria, line 73: lets add user to storage directly in case `register()` method works wrong

MaliukDaria (changes requested):

zachar-chorny (changes requested):
You didn't fix Masha's comment about checking a user for null. Do that and write an appropriate test

zachar-chorny (approved):

PR #283, opened 2022-08-22
solution

JirkVorl:


ira0807 (changes requested):

ira0807 (approved):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,105 @@
...
+    void register_nullPassword_NotOk() {
+        user.setPassword(null);
+        assertThrows(RuntimeException.class, () -> {
+            registrationService.register(user);
+        });
+    }
+
+    @Test
+    void register_validPasswordLength_NotOk() {
+        user.setPassword("123");
```
Noct2000, line 87: it would be better to use a wrong password that more closer to valid value

Noct2000 (approved):

PR #310, opened 2022-10-06
Registration service unit tests implementation v1

EvgenyTyshchenko:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,34 @@
...
+        User userByLogin = storageDao.get(user.getLogin());
+        if (userByLogin != null) {
+            if (userByLogin.equals(user)) {
+                throw new RuntimeException("User " + user.toString()
+                        + " already exists in the storage");
+            } else {
+                throw new RuntimeException("Login " + user.getLogin()
+                        + " is already taken by another user");
+            }
+        }
```
SoulARC, line 35: I think this method can be divided into several

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,135 @@
...
+import core.basesyntax.dao.StorageDao;
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceTest {
+    private static RegistrationService registrationService;
+    private static StorageDao storageDao;
+
```
SoulARC, line 15: Perhaps we should create a private user and use the @BeforeAll annotation to initialize it, this will reduce the amount of code.

SoulARC (changes requested):

SoulARC (approved):

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,133 @@
...
+        user.setAge(25);
+    }
+
+    @AfterEach
+    void clearStorage() {
+        Storage.people.clear();
+    }
+
+    @Test
+    void nullUser_NotOK() {
```
liliia-ponomarenko, line 40:
```suggestion
    void nullUser_notOK() {
```
the first letter after ```_``` in lowercase (check it everywhere)

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,133 @@
...
+    void invalidPassword_NotOK() {
+        user.setPassword("12345");
+        assertThrows(RuntimeException.class, () -> {
+            registrationService.register(user);
+        });
+    }
+
+    @Test
+    void registerAnotherUserWithSameLogin_NotOK() {
+        registrationService.register(user);
```
liliia-ponomarenko, line 96: You can’t use method register() here because you testing this method. You need to add a user to the base like this: Storage.people.add().
And then register the user with the same email using register() method expecting an exception. Check  it everywhere

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @BeforeEach
+    void setUp() {
+        user = new User();
+        user.setLogin("User login");
+        user.setPassword("User password");
+        user.setAge(25);
+    }
+
+    @AfterEach
```
liliia-ponomarenko, line 34: usually, this method is placed at the end test class

liliia-ponomarenko (changes requested):
Good job!🔥 Let's fix some mistakes😉

liliia-ponomarenko (approved):
great!🔥

PR #316, opened 2022-10-17
RegistrationService and tests implemented

hpavlo:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,110 @@
...
+        });
+    }
+
+    @Test
+    public void register_loginIsNull_notOk() {
+        User userWithLoginIsNull = createUser(null, USER_PASSWORD, USER_AGE);
+        assertThrows(RuntimeException.class, () -> {
+            registrationService.register(userWithLoginIsNull);
+        });
+    }
```
YuriiHryniv, line 34: I would additionally check whether the login is not empty, doesn't start with symbols, numbers,  doesn't have any uppercase letters, and contains "@" symbol

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Test
+    public void register_userWithCorrectData_ok() {
+        String uniqueLogin = "Alisa9987";
+        User normalUser = createUser(uniqueLogin, USER_PASSWORD, USER_AGE);
+        assertEquals(normalUser, registrationService.register(normalUser));
+    }
+
```
YuriiHryniv, line 102: I think it'a good idea to cover with tests user's id, equals and hashcode as well

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,110 @@
...
+    private User createUser(String login, String password, Integer age) {
+        User user = new User();
+        user.setLogin(login);
+        user.setPassword(password);
+        user.setAge(age);
+        return user;
+    }
+}
```
YuriiHryniv, line 110: I would move this logic to setUp method as Bohdan did in the video

YuriiHryniv (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Pavlo-Gook: I understand your point, but I don't see the need for it, since the task does not specify a check for a valid email like login.
I will wait for the mentor's comment ;)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Pavlo-Gook: I thought about it, but I use this method to create a new User object in each test and setUp method doesn't work for me

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Pavlo-Gook: We don't need to write the id value ourselves and I don't think it's worth adding tests for the equals and hashCode methods to work

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Pavlo-Gook: I will replace this method with a constructor in the User class

YuriiHryniv (approved):
I think it's good

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,49 @@
...
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new RuntimeException("User can't be null");
+        }
+
+        loginValidation(user.getLogin());
+        passwordValidation(user.getPassword());
+        ageValidation(user.getAge());
+
```
ChumakAn, line 21: remove emty lines

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,49 @@
...
+    private static final int MIN_PASSWORD_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new RuntimeException("User can't be null");
+        }
+
```
ChumakAn, line 17: remove empty line

ChumakAn (changes requested):
just empty lines)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
Pavlo-Gook: I thought empty lines improve the readability of the code)

ChumakAn (approved):

PR #324, opened 2022-10-19
implemented task

vov4ik89:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,25 @@
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int minPassLength = 6;
+    private static final int adultAge = 18;
```
anastasiavasylyk, line 9: It`s not a valid names for constant variables https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,25 @@
...
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new NullPointerException("Login can't be null");
+        }
```
anastasiavasylyk, line 16: Not a correct throw message if (user == null) it`s all object is null not only Login

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+        Storage.people.clear();
+    }
+
+    @Test
+    void register_nullLogin_NotOk() {
+        user = null;
+        assertThrows(NullPointerException.class, () -> {
+            registrationService.register(user);
+        });
+    }
```
anastasiavasylyk, line 35: here you check if all object user is null. you need to create separate method that will check all separate fields in user object on null and method that will check all object user on null

anastasiavasylyk (changes requested):
Let`s fix some mistakes

anastasiavasylyk (approved):
Approve for mentor

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,31 @@
...
+            throw new RuntimeException("It's already exists user by login " + user.getLogin());
+        }
+        if (user.getAge() <= ADULT_AGE) {
+            throw new RuntimeException("Can't register user with age " + user.getAge());
+        }
+        if (user.getPassword().length() < MIN_PASS_LENGTH) {
+            throw new RuntimeException("Password must contains at least "
+                    + MIN_PASS_LENGTH + " characters");
+        }
+        return storageDao.add(user);
```
ChumakAn, line 33: let's divide it in smaller methods for validating age, password and login

ChumakAn (changes requested):
Good , let's fix it

ChumakAn (approved):

PR #339, opened 2022-10-24
Added implementation of RegistrationService. 100% RegistrationService…

NSay0:
…Impl was covered with tests

### DIFF
--- a/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
+++ b/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
@@ -33,20 +33,15 @@ class RegistrationServiceImplTest {
         testUser.setAge(VALID_AGE);
     }
 
-    @AfterEach
-    void afterEach() {
-        Storage.people.clear();
-    }
-
     @Test
-    void register_userIsNull_NotOk() {
+    void register_userIsNull_notOk() {
         assertThrows(RuntimeException.class, () -> {
             registrationService.register(null);
         });
     }
 
     @Test
-    void register_userAgeIsNull_NotOk() {
+    void register_userAgeIsNull_notOk() {
         testUser.setAge(null);
         assertThrows(RuntimeException.class, () -> {
             registrationService.register(testUser);
@@ -54,7 +49,7 @@ class RegistrationServiceImplTest {
     }
 
     @Test
-    void register_userLoginIsNull_NotOk() {
+    void register_userLoginIsNull_notOk() {
         testUser.setLogin(null);
         assertThrows(RuntimeException.class, () -> {
             registrationService.register(testUser);
@@ -62,7 +57,7 @@ class RegistrationServiceImplTest {
     }
 
     @Test
-    void register_userPasswordIsNull_NotOk() {
+    void register_userPasswordIsNull_notOk() {
         testUser.setPassword(null);
         assertThrows(RuntimeException.class, () -> {
             registrationService.register(testUser);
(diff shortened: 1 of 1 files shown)


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @BeforeEach
+    void beforeEach() {
+        testUser = new User();
+        testUser.setLogin(DEFAULT_LOGIN);
+        testUser.setPassword(VALID_PASSWORD);
+        testUser.setAge(VALID_AGE);
+    }
+
+    @AfterEach
```
liliia-ponomarenko, line 36: this method is usually placed at the end of the classes

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+    void afterEach() {
+        Storage.people.clear();
+    }
+
+    @Test
+    void register_userIsNull_NotOk() {
```
liliia-ponomarenko, line 42:
```suggestion
    void register_userIsNull_notOk() {
```
every part of the name of the test method has to start with a letter in lowerCase ( check it everywhere )

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+    void register_userPasswordIsNull_NotOk() {
+        testUser.setPassword(null);
+        assertThrows(RuntimeException.class, () -> {
+            registrationService.register(testUser);
+        });
+    }
+
+    @Test
+    void register_withExistingLogin_NotOk() {
+        registrationService.register(testUser);
```
liliia-ponomarenko, line 74: You can’t use method register() here because you testing this method. You need to add a user to the base like this: Storage.people.add().
And then register the user with the same email using register() method expecting an exception. Check it everywhere

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+        assertThrows(RuntimeException.class, () -> {
+            registrationService.register(testUser);
+        });
+    }
+
+    @Test
+    void register_userAgeLessThen18_NotOk() {
```
liliia-ponomarenko, line 81: Don't use numbers in the names of test methods. better use valid/invalid

liliia-ponomarenko (changes requested):
good job! Let’s  improve your solution 😉

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
NSay0: moved to the end of class in new commit.
+correctec in new commit.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
NSay0: Renamed in new commit. L8r suffixes: *_NotOk and *_Ok was named just according to Bohdan video.  ¯\_(ツ)_/¯
+corrected in new commit.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
NSay0: +Renamed in new commit.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
NSay0: Ty, now it looks better. Only one such logic-misstake was found in tests.
+Corrected in new commit.

liliia-ponomarenko (approved):
great😎🔥

PR #361, opened 2022-10-26
Registration-validation test written.

luka-lozovyk:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,42 @@
...
+        checkLogin(user);
+        checkPassword(user);
+        return storageDao.add(user);
+    }
+
+    private void checkAge(User user) {
+        if (user.getAge() < ADULT_AGE) {
+            throw new RuntimeException("Can't register user age under 18");
+        }
+        if (user.getAge() >= ADULT_AGE) {
```
Mar4uk21, line 24: This is redundant because you add user 2 times.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,42 @@
...
+    private void checkLogin(User user) {
+        if (user.getLogin() == null) {
+            throw new RuntimeException("Login can't be NULL");
+        }
+        if (user.equals(storageDao.get(user.getLogin()))) {
+            throw new RuntimeException("User with such login already exists");
+        }
+    }
+
+    private void checkPassword(User user) {
```
Mar4uk21, line 38: You wrote a test that checks password is empty, but you haven't this check in  implementation, and you can the same check in checklogin method and write test about that.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Test
+    void checkStorageSizeAfterAddingTwoUsers() {
+        Storage.people.add(new User(1L,LOGIN, VALID_PASSWORD, VALID_AGE));
+        Storage.people.add(new User(1L,LOGIN_2, VALID_PASSWORD, VALID_AGE));
+        assertEquals(2, Storage.people.size());
+    }
+
+    @AfterEach
+    void tearDown() {
```
Mar4uk21, line 115: You can rename this method like clearStorage

Mar4uk21 (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
luka-lozovyk: We already have a limit of password length which is 6. Empty string is under that limit? so I think there is no need to check it.  But in login we need this check, because there is no length requrement of login.

Mar4uk21: Sorry, you're right

Mar4uk21 (approved):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Test
+    void registerUserWithExistingLogin_NotOk() {
+        Storage.people.add(defaultUser);
+        assertThrows(RuntimeException.class, () ->
+                registrationService.register(new User(22L,
+                        defaultUser.getLogin(),
+                        VALID_PASSWORD,
+                        VALID_AGE)));
```
liliia-ponomarenko, line 86: better firstly create a user and pass the user variable to assertThrows

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Test
+    void register_userWithNullPassword() {
+        defaultUser.setPassword(null);
+        assertThrows(RuntimeException.class, () -> registrationService.register(defaultUser));
+    }
+
+    @Test
+    void register_userWithValidPassword_ok() {
+        assertTrue(Storage.people.add(defaultUser));
```
liliia-ponomarenko, line 97: how is this test related to the testing method ```register()```?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,124 @@
...
+        defaultUser.setPassword(EMPTY_LINE);
+        assertThrows(RuntimeException.class, () ->
+                registrationService.register(defaultUser));
+    }
+
+    @Test
+    void checkStorageSizeAfterAddingTwoUsers() {
+        Storage.people.add(new User(1L,LOGIN, VALID_PASSWORD, VALID_AGE));
+        Storage.people.add(new User(1L,LOGIN_2, VALID_PASSWORD, VALID_AGE));
+        assertEquals(2, Storage.people.size());
```
liliia-ponomarenko, line 117: the same about this test

liliia-ponomarenko (changes requested):
Let’s  fix a few mistakes 😉

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,124 @@
...
+    @Test
+    void register_userWithEmptyLinePassword_notOk() {
+        defaultUser.setPassword(EMPTY_LINE);
+        assertThrows(RuntimeException.class, () -> registrationService.register(defaultUser));
+    }
+
+    @Test
+    void checkStorageSizeAfterAddingTwoUsers() {
+        User validUser1 = new User(LOGIN, VALID_PASSWORD, VALID_AGE);
+        User validUser2 = new User(LOGIN_2, VALID_PASSWORD, VALID_AGE);
```
liliia-ponomarenko, line 114: Don’t use counters for variable names. it’s  bad practice

liliia-ponomarenko (approved):
good job! 🔥

PR #382, opened 2022-11-23
implementation RegistrationService and RegistrationServiceTest

BusyhinD:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,24 @@
...
+    private static final int MINIMUM_USER_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new RuntimeException("User cannot be null!");
+        }
+        if (user.getPassword().length() < MINIMUM_PASSWORD_LENGTH) {
```
aaananas, line 17: what if password is null, this would produce NullPointerException
this exception is never to be tested - it should not appear

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Test
+    void register_validUser_Ok() {
+        assertEquals(user, registrationService.register(user));
+        assertTrue(Storage.people.contains(user));
+    }
+
+    @Test
+    void register_nullLogin_Ok() {
+        user.setLogin(null);
```
Rommelua, line 97: Why is it ok???

Rommelua (approved):

PR #393, opened 2022-12-12
first made tests

Sanderfuck:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @AfterEach
+    void clearStorage() {
+        Storage.people.clear();
+    }
+
+    @Test
+    void registerUsersWithInvalidCredentials_notOk() {
+        assertThrows(ValidationException.class, () -> registrationService.register(userTwo));
+        assertThrows(ValidationException.class, () -> registrationService.register(userThree));
+        assertThrows(ValidationException.class, () -> registrationService.register(nullUser));
```
nacenik, line 60: Also, don't create test methods, that test different scenarios. One test per scenario.

For example: if you cannot create a new user because of his age - make it in one test.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+    void registerWithValidAndInvalidCredentialsUsers() {
+        assertEquals(userOne, registrationService.register(userOne));
+        assertThrows(ValidationException.class, () -> registrationService.register(userOne));
+        assertThrows(ValidationException.class, () -> registrationService.register(nullUser));
+        assertThrows(ValidationException.class, () -> registrationService.register(userTwo));
+        assertEquals(userFour, registrationService.register(userFour));
+    }
+
+    @Test
+    void registerWithSameUser_notOk() {
```
nacenik, line 74: Please, rename this method according to the template in the task description.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+        assertEquals(userOne, registrationService.register(userOne));
+        assertThrows(ValidationException.class, () -> registrationService.register(userOne));
+    }
+
+    @Test
+    void register_Ok() {
```
nacenik, line 80: Please, rename this method according to the template in the task description.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+        assertEquals(userOne, registrationService.register(userOne));
+        assertThrows(ValidationException.class, () -> registrationService.register(userOne));
+        assertThrows(ValidationException.class, () -> registrationService.register(nullUser));
+        assertThrows(ValidationException.class, () -> registrationService.register(userTwo));
+        assertEquals(userFour, registrationService.register(userFour));
+    }
+
+    @Test
+    void registerWithSameUser_notOk() {
+        assertEquals(userOne, registrationService.register(userOne));
```
nacenik, line 75: Use storageDao.add instead of register method to add user for testing.

nacenik (changes requested):
Please, fix all described problem

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,38 @@
...
+        if (user == null) {
+            throw new ValidationException("User can`t be null");
+        }
+
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new ValidationException("User with this login exist in storage");
+        }
+
+        if (user.getAge() < MIN_USERS_AGE) {
+            throw new ValidationException("User`s age can`t be less than 18");
```
okuzan, line 31: Please refer to the constant you declared earlier. You can forget to change the hardcoded value when the requirements change, so having to change the value only in one place (constant) is a less error-prone approach.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,38 @@
...
+        }
+
+        if (user.getPassword().length() < MIN_PASSWORD_LENGTH) {
+            throw new ValidationException("User`s password length can`t be less than 6");
```
okuzan, line 35: Please refer to the constant

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,38 @@
...
+    private static final int MIN_PASSWORD_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (validation(user)) {
+            return storageDao.add(user);
+        }
+        throw new ValidationException("This user can`t be registered");
```
okuzan, line 18: I don't think we gain something by throwing ValidationException twice, all needed information will be in the first exception

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static RegistrationServiceImpl registrationService;
+    private static User userOne;
+    private static User userTwo;
+    private static User userThree;
+    private static User nullUser;
+    private static User userFour;
```
okuzan, line 19: naming with numbers is a bad practice. Instead, try to use the name to describe its holder - like, `validUser`, `badAgeUser` etc.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+        assertEquals(userOne, registrationService.register(userOne));
+        assertThrows(ValidationException.class, () -> registrationService.register(userOne));
+    }
+
+    @Test
+    void register_userWithValidCredentials_Ok() {
+        assertEquals(userOne, registrationService.register(userOne));
+        assertEquals(userFour, registrationService.register(userFour));
+    }
+}
```
okuzan, line 78: There are more scenarios to test: null/empty password, null/empty login, negative/zero age etc

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
okuzan, line 19: tests should be isolated from each other, but static members belong to the whole class. A better approach would be to declare and configure a test user object inside each method as a local variable

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Test
+    void register_UsersWithInvalidCredentials_notOk() {
+        assertThrows(ValidationException.class, () -> registrationService.register(userTwo));
+        assertThrows(ValidationException.class, () -> registrationService.register(userThree));
+    }
+
+    @Test
+    void register_withFewSameUser_notOk() {
```
okuzan, line 68: a bit confusing naming, maybe try `register_existingUser_notOk()`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,38 @@
...
+        }
+
+        if (user.getAge() < MIN_USERS_AGE) {
+            throw new ValidationException("User`s age can`t be less than 18");
+        }
+
+        if (user.getPassword().length() < MIN_PASSWORD_LENGTH) {
+            throw new ValidationException("User`s password length can`t be less than 6");
+        }
+        return true;
```
okuzan, line 37: also check if age is not null (it's object, not primitive), if password and login are not null/empty

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,38 @@
...
+
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new ValidationException("User with this login exist in storage");
+        }
+
+        if (user.getAge() < MIN_USERS_AGE) {
+            throw new ValidationException("User`s age can`t be less than 18");
+        }
+
+        if (user.getPassword().length() < MIN_PASSWORD_LENGTH) {
```
okuzan, line 34: this could throw NPE if password isn't checked first

okuzan (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,47 @@
...
+        if (user.getPassword() == null || user.getPassword().length() < MIN_PASSWORD_LENGTH) {
+            throw new ValidationException("User`s password length can`t be null or less than "
+                    + MIN_PASSWORD_LENGTH);
+        }
+
+        if (user.getAge() == null || user.getAge() < MIN_USERS_AGE) {
+            throw new ValidationException("User`s age can`t be null or less than " + MIN_USERS_AGE);
+        }
+
+        if (user.getLogin() == null) {
```
okuzan, line 39: could also add isEmpty() check

okuzan (approved):
Nicely done!

PR #400, opened 2022-12-18
1

dimonkoval:


src/main/java/core/basesyntax/exceptions/ValidationException.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax.exceptions;
+
+public class ValidationException extends Exception {
```
zachar-chorny, line 3: If you want to use your custom exception,extend it from RuntimeException

src/main/java/core/basesyntax/service/RegistrationService.java
```diff
@@ -1,7 +1,8 @@
 package core.basesyntax.service;
 
+import core.basesyntax.exceptions.ValidationException;
 import core.basesyntax.model.User;
 
 public interface RegistrationService {
-    User register(User user);
+    User register(User user) throws ValidationException;
```
zachar-chorny, line 7: What do you think should be done using your method? In this case you don't need to catch anything, just send incorrect data and throw RuntimeException
```suggestion
    User register(User user);
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exceptions.ValidationException;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
```
zachar-chorny, line 13: One positive test will be enough

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+        user.setAge(19);
+        registration.register(user);
+        assertEquals(user, storageDao.get("Don"));
+    }
+
+    private void checkAssert(User user) {
+        assertThrows(ValidationException.class, () -> {
+            registration.register(user);
+        });
+    }
```
zachar-chorny, line 106: This is usually good practice but not in tests. It is better when the assert method is in the test itself, because otherwise it is hard to read

zachar-chorny (changes requested):

aleksandr-bevza (approved):

PR #410, opened 2022-12-24
done

olei23:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Test
+    void register_returnUserId() {
+        Long id = registrationService.register(testUser).getId();
+        assertNotNull(id);
+    }
+
+    @Test
+    void register_addingValidUser_Ok() {
```
karma-o, line 90:
```suggestion
    void register_ValidUser_Ok() {
```

karma-o (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,53 @@
...
+        }
+        if (user.getLogin() == null) {
+            throw new RegistrationException("Login can't be null");
+        }
+        if (user.getAge() == null) {
+            throw new RegistrationException("Age can't be null");
+        }
+        if (user.getPassword() == null) {
+            throw new RegistrationException("Password can't be null");
+        }
```
karma-o, line 32: better to move null checks to their exclusive methods, so that `checkPassword` for example would check if it is not null and valid from the requirements point of view so that this method would be solely responsible for password validation.
Same for the other propertioes

karma-o (approved):
left a comment and approved assuming you will fix it :wink:

PR #439, opened 2023-01-06
jv-registration-validation.

Stim1:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,71 @@
...
+            throw new UserException("User field can't be empty.");
+        }
+        userLogin(user.getLogin());
+        userPassword(user.getPassword());
+        userAge(user.getAge());
+        userId(user.getId());
+        return storageDao.add(user);
+    }
+
+    private void userLogin(String login) {
```
kozhukhovsky, line 25:
```suggestion
    private void checkUserLogin(String login) {
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,71 @@
...
+        if (login.isEmpty()) {
+            throw new UserException("Login field can't be empty.");
+        }
+        if (storageDao.get(login) != null) {
+            throw new UserException("Login: " + login
+                    + " is already exist, please choose another login.");
+        }
+    }
+
+    private void userPassword(String password) {
```
kozhukhovsky, line 38:
```suggestion
    private void checkUserPassword(String password) {
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,71 @@
...
+        if (password.isEmpty()) {
+            throw new UserException("Password field can't be empty.");
+        }
+        if (password.length() < MIN_PASSWORD_LENGTH) {
+            throw new UserException("The password length must be minimum " + MIN_PASSWORD_LENGTH
+                    + " characters, and your password length is : " + password.length());
+        }
+    }
+
+    private void userAge(Integer age) {
```
kozhukhovsky, line 57:
```suggestion
    private void checkUserAge(Integer age) {
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,71 @@
...
+            throw new UserException("To register ,your age must be at least: "
+                    + MIN_USER_AGE + " ,instead, your age is: " + age);
+        }
+    }
+
+    private void userId(Long id) {
+        if (id == null) {
+            throw new UserException("Id is null.");
+        }
     }
```
kozhukhovsky, line 71:
```suggestion
```

Usually the ID is given after registration by the database, there is no point in checking now

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Test
+    void register_userValid_Ok() {
+        registrationService.register(user);
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
+
+    }
+
+    @Test
+    void register_login_isNull_notOk() {
```
kozhukhovsky, line 43: login_isNull, but ageNull

Use the same method naming convention

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,65 @@
...
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_USER_AGE = 18;
+    private static final int MIN_PASSWORD_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new InvalidUserException("User field can't be empty.");
```
karma-o, line 16: user is not a field here, it is a parameter and it is not empty, it is null, empty is not the same as null, so we should be careful with those words
message can be like: `User cant be null`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,65 @@
...
+    private void checkUserPassword(String password) {
+        if (password == null) {
+            throw new InvalidUserException("Password field is null.");
+        }
+        if (password.contains(password.toUpperCase())) {
+            throw new InvalidUserException("The password cannot contain only uppercase letters");
+        }
+        if (password.contains(password.toLowerCase())) {
+            throw new InvalidUserException("The password cannot contain only lowercase letters");
+        }
```
karma-o, line 46: there is nothing about that in the task description, we should not invent some additional logic, being a developer is more about implementing the requirements of the stakeholders

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,65 @@
...
+        if (password.isEmpty()) {
+            throw new InvalidUserException("Password field can't be empty.");
+        }
+        if (password.length() < MIN_PASSWORD_LENGTH) {
+            throw new InvalidUserException("The password length must be minimum "
+                    + MIN_PASSWORD_LENGTH
+                    + " characters, and your password length is : " + password.length());
+        }
```
karma-o, line 54: we can just remove the empty check since we already have a min length check

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,101 @@
...
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @BeforeEach
+    void setUp() {
+        user = new User();
+        user.setLogin("mate_user");
+        user.setAge(18);
+        user.setPassword("MyCatNameDin");
+        user.setId(713298L);
```
karma-o, line 27: we should not have the id of the user here, since he is not registered yet, our storage is assigning ids to users after registration

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @AfterEach
+    void removeStorage() {
+        Storage.people.clear();
+    }
+
+    @Test
+    void register_userValid_Ok() {
```
karma-o, line 36: all of your tests are just exact duplicates, changing their name doesn't mean that they will work in another way. 
You should not have an exception thrown when you are passing a valid user, that is the whole point of the tests.
In `ok` test case you should check whether the user `id` has been set.
And you should change test user fields according to your test name, e.g. 
```
@Test
    void register_loginNull_notOk() {
        user.setLogin(null);
        assertThrows(RuntimeException.class, () -> registrationService.register(user));
    }
```
this will be the same user that you have initialized in the `setUp()` method but with null login

karma-o (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+    @BeforeEach
+    void defaultUser() {
+        user = new User();
+        user.setLogin("stim_one");
+        user.setAge(MIN_AGE);
+        user.setPassword("MyCatNameDin");
+    }
+
+    @AfterEach
+    void removeStorage() {
```
Rommelua, line 31:
```suggestion
    void clearStorage() {
```

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+        Storage.people.clear();
+    }
+
+    @Test
+    void register_userValid_Ok() {
+        registrationService.register(user);
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
```
Rommelua, line 38: in every test you should expect **InvalidUserException** (not RuntimeException)

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Test
+    void register_loginNull_notOk() {
+        user.setLogin(null);
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void register_age_minimumAge_notOk() {
```
Rommelua, line 49:
```suggestion
    void register_lowAge_notOk() {
```

Rommelua (changes requested):

kozhukhovsky (approved):

PR #442, opened 2023-01-12
Registration validation solution

nkharyshyn:
Created custom unchecked InvalidDataException and threw it in register method.
Wrote logic for register method to comply with all conditions. Added constructor for User class.
Created:
1) BeforeAll to initialize variables and BeforeEach to clear list before each test 
2) test to register valid user, check if list contains them after registration and check if list size is equal to amount of registered user 
3) test to check if it is not possible to register new user if his any data is invalid 
4) test to check if it is not possible to register null user

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -2,13 +2,47 @@ package core.basesyntax.service;
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exception.InvalidDataException;
 import core.basesyntax.model.User;
 
-public class RegistrationServiceImpl implements RegistrationService {
+public class RegistrationServiceImpl
+        implements RegistrationService {
+    private static final int OLDEST_HUMAN_ALIVE = 118;
+    private static final int MIN_AGE = 18;
+    private static final int ZERO_AGE = 0;
+    private static final int MIN_PASSWORD_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new InvalidDataException("User can't be null");
+        }
+        if (user.getLogin() == null) {
+            throw new InvalidDataException("Login can't be null");
+        }
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidDataException("User with such login already exists");
+        }
+        if (user.getPassword() == null) {
+            throw new InvalidDataException("Password can't be null");
+        }
+        if (user.getPassword().length() < MIN_PASSWORD_LENGTH) {
+            throw new InvalidDataException("Password must be at least "
+                    + MIN_PASSWORD_LENGTH + " characters");
+        }
+        if (user.getAge() == null) {
+            throw new InvalidDataException("Age can't be null");
+        }
+        if (user.getAge() < ZERO_AGE) {
+            throw new InvalidDataException("User's age cannot be less than " + ZERO_AGE);
+        }
+        if (user.getAge() < MIN_AGE) {
+            throw new InvalidDataException("User must be at least " + MIN_AGE + " years old");
+        }
+        if (user.getAge() > OLDEST_HUMAN_ALIVE) {
+            throw new InvalidDataException("Incorrect age. Please enter your real age");
+        }
+        return storageDao.add(user);
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,47 @@
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exception.InvalidDataException;
 import core.basesyntax.model.User;
 
-public class RegistrationServiceImpl implements RegistrationService {
+public class RegistrationServiceImpl
+        implements RegistrationService {
+    private static final int OLDEST_HUMAN_ALIVE = 118;
```
nacenik, line 10:
```suggestion
    private static final int MAX_AGE = 118;
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,47 @@
...
+            throw new InvalidDataException("Password can't be null");
+        }
+        if (user.getPassword().length() < MIN_PASSWORD_LENGTH) {
+            throw new InvalidDataException("Password must be at least "
+                    + MIN_PASSWORD_LENGTH + " characters");
+        }
+        if (user.getAge() == null) {
+            throw new InvalidDataException("Age can't be null");
+        }
+        if (user.getAge() < ZERO_AGE) {
```
nacenik, line 37: this condition is unnecessary. Users with ages less than 18 don't allow to register.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    @BeforeEach
+    void setUp() {
+        Storage.people.clear();
+        size = 0;
+    }
+
+    @Test
+    void register_newUser_Ok() {
```
nacenik, line 58: Tons of redundant logic. You should check only RegistrationServiceImplementation behavior. Not Storage.people logic

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,160 @@
...
+    @Test
+    void register_incorrectPassword_notOk() {
+        assertThrows(InvalidDataException.class, () -> {
+            registrationService.register(shortPasswordUser);
+        }, "Test failed! If password length is invalid,"
+                + " we should throw custom unchecked exception");
+    }
+
+    @Test
+    void register_incorrectAge_notOk() {
```
nacenik, line 126: segregate this test on smaller parts.
Example:
```
@Test
     void register_veryOldUser_notOk() {
         assertThrows(InvalidDataException.class, () -> {
             registrationService.register(veryOldUser);
         }, "Test failed! If age is invalid,"
                 + " we should throw custom unchecked exception");
    }
....
```
One scenario - one test

nacenik (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,43 @@
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exception.InvalidDataException;
 import core.basesyntax.model.User;
 
-public class RegistrationServiceImpl implements RegistrationService {
+public class RegistrationServiceImpl
+        implements RegistrationService {
```
Ivan95kos, line 9:
```suggestion
public class RegistrationServiceImpl implements RegistrationService {
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Ivan95kos: Not fixed

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,43 @@
...
+        }
+        if (user.getAge() == null) {
+            throw new InvalidDataException("Age can't be null");
+        }
+        if (user.getAge() < MIN_AGE) {
+            throw new InvalidDataException("User must be at least " + MIN_AGE + " years old");
+        }
+        if (user.getAge() > MAX_AGE) {
+            throw new InvalidDataException("Incorrect age. Please enter your real age");
+        }
```
Ivan95kos, line 41: Hard to read.
It is better to divide into methods and describe the necessary logic in them.
private checkAge(User user) {
// age verification
}

Ivan95kos (changes requested):

Ivan95kos (approved):
Great work

PR #452, opened 2023-01-25
solved task

yaaariiik:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @BeforeEach
+    void setUp() {
+        User arnold = new User();
+        arnold.setAge(CORRECT_USERS_AGE);
+        arnold.setLogin("Arny");
+        arnold.setPassword("1*hD#Sak");
+        storageDao.add(arnold);
+    }
```
Ivan95kos, line 37: it is not necessary to perform this action before each test, 
as far as I understand it is needed only for this test "register_sameLogin_notOk()", 
then you only need to write it there

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,30 @@
...
+                    + MIN_USERS_AGE + " and " + MAX_USERS_AGE + ".");
+        }
+        if (user.getPassword() == null
+                || user.getPassword().length() < MIN_PASSWORD_LENGTH) {
+            throw new RegistrationException("Password length must be at least "
+                    + MIN_PASSWORD_LENGTH + " characters.");
+        }
+        if (user.getLogin() == null || storageDao.get(user.getLogin()) != null) {
+            throw new RegistrationException("User with this login is already exists. ");
+        }
```
Ivan95kos, line 31: It is better to take the validation data into private methods and call them here

src/main/java/core/basesyntax/service/RegistrationException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+public class RegistrationException extends RuntimeException {
+    public RegistrationException(String message) {
+        super("User not registered. " + message + " Please, check input data and try again.");
```
Ivan95kos, line 5: It is bad practice to change the message in the exception, it is better to output what is passed in the message

Ivan95kos (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,34 @@
...
+                    + MIN_USERS_AGE + " and " + MAX_USERS_AGE + ".");
+        }
+        if (user.getPassword() == null
+                || user.getPassword().length() < MIN_PASSWORD_LENGTH) {
+            throw new RegistrationException("Password length must be at least "
+                    + MIN_PASSWORD_LENGTH + " characters.");
+        }
+        if (user.getLogin() == null || storageDao.get(user.getLogin()) != null) {
+            throw new RegistrationException("User with this login is already exists.");
+        }
```
aleksandr-bevza, line 36: It would be easier to read your code if you create private methods for these checks:
```suggestion
        checkAge(user.getAge());
        checkPassword(user.getPassword());
        checkLogin(user.getLogin());       
```

aleksandr-bevza (changes requested):

aleksandr-bevza (approved):

PR #469, opened 2023-02-12
my solution jv-regisration-service

yaroslav-ometiukh:
-created custom exceptions for incorrect/null value;
-created tests;
-wrote register method according to restrictions;

### DIFF
--- a/src/main/java/core/basesyntax/customexception/ValidationExceptionIncorrectValue.java
+++ b/src/main/java/core/basesyntax/customexception/ValidationExceptionIncorrectValue.java
@@ -1,0 +1,7 @@
+package core.basesyntax.customexception;
+
+public class ValidationExceptionIncorrectValue extends RuntimeException {
+    public ValidationExceptionIncorrectValue(String errorMessage) {
+        super(errorMessage);
+    }
+}
--- a/src/main/java/core/basesyntax/customexception/ValidationExceptionNullValue.java
+++ b/src/main/java/core/basesyntax/customexception/ValidationExceptionNullValue.java
@@ -1,0 +1,7 @@
+package core.basesyntax.customexception;
+
+public class ValidationExceptionNullValue extends RuntimeException {
+    public ValidationExceptionNullValue(String errorMessage) {
+        super(errorMessage);
+    }
+}
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -1,14 +1,60 @@
 package core.basesyntax.service;
 
+import core.basesyntax.customexception.ValidationExceptionIncorrectValue;
+import core.basesyntax.customexception.ValidationExceptionNullValue;
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int USER_MIN_AGE = 18;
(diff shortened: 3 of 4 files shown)


src/main/java/core/basesyntax/customexception/ValidationExceptionIncorrectValue.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.customexception;
+
+public class ValidationExceptionIncorrectValue extends RuntimeException {
```
nacenik, line 3: every custom exception should end on word "Exception"
```suggestion
public class UserValidationException extends RuntimeException {
```

src/main/java/core/basesyntax/customexception/ValidationExceptionNullValue.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.customexception;
+
+public class ValidationExceptionNullValue extends RuntimeException {
+    public ValidationExceptionNullValue(String errorMessage) {
+        super(errorMessage);
+    }
+}
```
nacenik, line 7: delete this class. That's enough to create only one exception here

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @BeforeAll
+    static void setUp() {
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @BeforeEach
+    void setDefaultData() {
+        testUser = new User();
+        Storage.people.clear();
```
nacenik, line 26: to clean storage better use `@AfterEach`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    @Test
+    void containsUserCorrectData_OK() {
+        testUser.setLogin("CorrectLogin");
+        testUser.setAge(18);
+        testUser.setPassword("CorrectPassword");
```
nacenik, line 33: you can set all field for `testUser` in `@BeforeEach`and change some of them for every test to avoid duplicated code

nacenik (changes requested):

aleksandr-bevza (changes requested):
There is a chackstyle error. Fix it please
image

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,58 @@
...
+        } else if (!isLoginValid(user)) {
+            throw new UserValidationException(
+                    "Can't be empty or contain whitespace");
+        } else if (!isAgeValid(user)) {
+            throw new UserValidationException(
+                    "Must meet the minimum age requirement");
+        } else if (!isPasswordValid(user)) {
+            throw new UserValidationException(
+                    "Must be at least 6 characters without whitespace");
+        }
```
Ivan95kos, line 29: It is better to check and throw exceptions in the appropriate method, it will be more readable
```suggestion
        checkForNullValues(user);
        loginValid(user);
        userExists(user);
        ageValid(user);
        passwordValid(user);
```

Ivan95kos (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
yaroslav-ometiukh: Understood, thanks!

aleksandr-bevza (approved):

PR #480, opened 2023-02-17
Registration-validation solution

serhii-shyian:


src/main/java/core/basesyntax/model/User.java
```diff
@@ -40,6 +40,15 @@ public void setAge(Integer age) {
...
 
+    public User createUser(long id, String login, String password, int age) {
+        User user = new User();
+        user.setId(id);
+        user.setLogin(login);
+        user.setPassword(password);
+        user.setAge(age);
+        return user;
+    }
+
```
okuzan, line 51: don't change model. Moreover, it looks like you wanted to create a constructor

src/test/java/core/basesyntax/dao/StorageDaoImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class StorageDaoImplTest {
```
okuzan, line 15: we don't test dao in this task, delete this class

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,211 @@
...
+        assertThrows(RuntimeException.class, () -> registrationService.register(user),
+                "There must RuntimeException throw "
+                        + "when try REGISTER SAME User");
+        assertEquals(expectedSize, Storage.people.size(),
+                "Storage size not changed");
+    }
+
+    @Test
+    void registering_existentLoginUser_notOK() {
+        User[] users = new User[2];
```
okuzan, line 110: why do you need this array? you only use instances, you can do it directly

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,211 @@
...
+        assertEquals(actual, users[0], "Registration method "
+                + "must return registered User");
+        assertNotEquals(oldId, users[0].getId(), "User Id must should be changed "
+                + "due to registration");
+        expectedSize++;
+
+        assertThrows(RuntimeException.class, () -> registrationService.register(users[1]),
+                "There must RuntimeException throw "
+                        + "when try register User with SAME LOGIN");
+        assertEquals(expectedSize, Storage.people.size(), "Storage size not changed properly");
```
okuzan, line 126: you have way too much setup and asserts in all of your test methods. For example, this scenario is adequately tested by just two lines something like this (if your default user is setup with valid values): 
        registrationService.register(user);
        assertThrows(InvalidUserException.class, () ->
                registrationService.register(user), "User already exists!"

okuzan (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,38 @@
...
+        }
+        if (user.getAge() == null) {
+            throw new RegistrationException("User age field "
+                    + "cannot be empty");
+        }
+        if (user.getAge() < VALID_USER_AGE) {
+            throw new RegistrationException("Users age must be more "
+                    + VALID_USER_AGE
+                    + " years");
+        }
```
Ivan95kos, line 39: For the future, It is better to divide into methods and describe the necessary logic in them.
private checkAge(User user) {
// age verification
}

Ivan95kos (approved):
Great job

PR #482, opened 2023-02-20
Method register and test was implemented

GalubaGSV:


mishok321 (approved):
good job

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,34 @@
...
+        }
+        if (user.getLogin() == null || user.getLogin().contains(" ")) {
+            throw new RegisterValidationException("You enter wrong login");
+        }
+        if (user.getPassword() == null || user.getPassword().length() < 8) {
+            throw new RegisterValidationException("You enter wrong password");
+        }
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new RegisterValidationException("Login already exist!");
+        }
```
petro-ushchuk, line 31: Create private method `validate`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,34 @@
...
+                || user.getAge() < AGE_LOW_LIMIT
+                || user.getAge() > AGE_UP_LIMIT) {
+            throw new RegisterValidationException("You enter wrong age");
+        }
+        if (user.getLogin() == null || user.getLogin().contains(" ")) {
+            throw new RegisterValidationException("You enter wrong login");
+        }
+        if (user.getPassword() == null || user.getPassword().length() < 8) {
+            throw new RegisterValidationException("You enter wrong password");
+        }
```
petro-ushchuk, line 28: More details. `Length should be more than 8 symbols`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,34 @@
...
-        return null;
+        if (user == null) {
+            throw new RegisterValidationException("User must be Not null");
+        }
+        if (user.getAge() == null
+                || user.getAge() < AGE_LOW_LIMIT
+                || user.getAge() > AGE_UP_LIMIT) {
+            throw new RegisterValidationException("You enter wrong age");
+        }
+        if (user.getLogin() == null || user.getLogin().contains(" ")) {
```
petro-ushchuk, line 23: Extract `" "` to the constant.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,34 @@
...
+            throw new RegisterValidationException("You enter wrong login");
```
petro-ushchuk, line 24: Also, add detailed explanation why login is wrong.

petro-ushchuk (changes requested):
Good work, but there are a couple things we can improve!

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,44 @@
...
+                || user.getAge() < AGE_LOW_LIMIT
+                || user.getAge() > AGE_UP_LIMIT) {
+            throw new RegisterValidationException("You enter wrong age. Expected between "
+                    + AGE_LOW_LIMIT + " and " + AGE_UP_LIMIT + " but was " + user.getAge());
+        }
+        if (user.getLogin() == null || user.getLogin().contains(SPACE)) {
+            throw new RegisterValidationException("Your login must not contain a space");
+        }
+        if (user.getPassword() == null || user.getPassword().length() < 8) {
+            throw new RegisterValidationException("You password must be more then 8 character");
```
karma-o, line 39: it is always nice to specify the inclusivity, looks like not that important but adds a lot more clarity
```suggestion
            throw new RegisterValidationException("You password must be at least 8 characters long");
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,44 @@
...
+        }
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new RegisterValidationException("Login already exist!");
```
karma-o, line 42:
```suggestion
            throw new RegisterValidationException("user with such login already exists!");
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+        Storage.people.clear();
+    }
+
+    @Test
+    void register_nullUser_NotOk() {
+        testUser = null;
+        assertThrows(RegisterValidationException.class, () -> {
+            registrationService.register(testUser);
+        }, "User must be NOT null");
```
karma-o, line 66: I think you misunderstood how the messages in assertThrows(or any assert really) work, this message is printed when the test is failed so they should represent some info on why exactly our test failed in this case it can be:
`Expected " + RegistrationValidationException.class.getName() + " to be thrown for the null user, but it wasn't`
Check all of the other messages

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Test
+    void register_less18YearsAge_NotOk() {
+        testUser.setAge(AGE_LES_18_YEARS);
+        assertThrows(RegisterValidationException.class, () -> {
+            registrationService.register(testUser);
+        }, "Age must be more then 18");
```
karma-o, line 74: another message example:
`register should not save users under 18 years old, but it did`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+    void register_password_LessThen_8_NotOk() {
+        testUser.setPassword(PASSWORD_8_SYMBOL_NOTOK);
+        assertThrows(RegisterValidationException.class, () -> {
+            registrationService.register(testUser);
+        }, "Password must be mo then 8 characters");
+    }
+
+    @Test
+    void register_login_OlreadyExist_NotOk() {
+        registrationService.register(testUser);
```
karma-o, line 127: In each test case when you test some behavior with some already existing user, we should not put him there using the same register method which is under test right now, and you cannot guarantee that it works as intended.
Instead, we should put this test data directly to `Storage` like `Storage.add(user)` or as close to the destination as possible, the lower the better.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+        testUser.setLogin(LOGIN_OK);
+        testUser.setPassword(PASSWORD_8_SYMBOL_OK);
+
+        secondTestUser = new User();
+        secondTestUser.setId(SECOND_TEST_ID);
+        secondTestUser.setAge(AGE_MORE_18_YEARS);
+        secondTestUser.setLogin(LOGIN_SECOND_OK);
+        secondTestUser.setPassword(PASSWORD_8_SYMBOL_SECOND_OK);
+
+        Storage.people.clear();
```
karma-o, line 58: I'd rather move this to the after each method

karma-o (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,177 @@
...
+    private static final long SECOND_TEST_ID = 25978;
+
+    private static final Integer AGE_18_YEARS = 18;
+    private static final Integer AGE_LES_18_YEARS = 17;
+    private static final Integer AGE_MORE_18_YEARS = 19;
+    private static final Integer AGE_MORE_80_YEARS = 83;
+
+    private static final String LOGIN_OK = "GrebanaRusnya";
+    private static final String LOGIN_SECOND_OK = "PutinLoh";
+    private static final String LOGIN_CONTAINS_SPACE = "Slava Ukraine";
```
petro-ushchuk, line 28: 🇺🇦

petro-ushchuk (approved):
Good work! 🔥

PR #483, opened 2023-02-21
I implement method register in class RegistrationServiceImpl and writ…

NazarZva:
…e tests to this implementation

### DIFF
--- a/src/main/java/core/basesyntax/model/User.java
+++ b/src/main/java/core/basesyntax/model/User.java
@@ -8,6 +8,13 @@ public class User {
     private String password;
     private Integer age;
 
+    public User(Long id, String login, String password, Integer age) {
+        this.id = id;
+        this.login = login;
+        this.password = password;
+        this.age = age;
+    }
+
     public Long getId() {
         return id;
     }
--- a/src/main/java/core/basesyntax/service/RegistrationService.java
+++ b/src/main/java/core/basesyntax/service/RegistrationService.java
@@ -3,5 +3,5 @@ package core.basesyntax.service;
 import core.basesyntax.model.User;
 
 public interface RegistrationService {
-    User register(User user);
+    User register(User user) throws UserIsNotValidException;
 }
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -2,13 +2,27 @@ package core.basesyntax.service;
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+//import core.basesyntax.db.Storage;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
-    public User register(User user) {
-        return null;
+    public User register(User user) throws UserIsNotValidException {
+        if (user == null) {
+            throw new UserIsNotValidException("user is null");
+        }
+        if (user.getLogin() == null || user.getPassword() == null) {
+            throw new UserIsNotValidException("users login or password is null");
+        }
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new UserIsNotValidException("user is already registered");
+        }
+        if (user.getAge() < 18 || user.getPassword().length() < 6) {
+            throw new UserIsNotValidException("incorrect password or age");
+        }
+        storageDao.add(user);
+        return user;
     }
 }
--- a/src/main/java/core/basesyntax/service/UserIsNotValidException.java
+++ b/src/main/java/core/basesyntax/service/UserIsNotValidException.java
@@ -1,0 +1,7 @@
+package core.basesyntax.service;
+
+public class UserIsNotValidException extends RuntimeException {
+    public UserIsNotValidException(String message) {
+        super(message);
+    }
+}
--- a/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
+++ b/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
@@ -1,0 +1,132 @@
+package core.basesyntax.service;
+
+import core.basesyntax.dao.StorageDao;
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class RegistrationServiceImplTest {
+    private static RegistrationService registrationService;
+    private static StorageDao storage;
+    private User user1;
+    private User user2;
+    private User user3;
+
+    @BeforeAll
+    static void beforeAll() {
+        registrationService = new RegistrationServiceImpl();
+        storage = new StorageDaoImpl();
+    }
+
+    @BeforeEach
+    void setUp() {
+        user1 = new User(1L,"login1", "password1", 22);
+        user2 = new User(2L,"login2", "password2", 20);
+        user3 = new User(3L,"login3", "password3", 18);
+    }
+
+    @Test
+    void registerAndGetOrdinaryUsers() {
+        registrationService.register(user1);
+        User actual = storage.get("login1");
+        assertEquals(user1,actual);
+
+        registrationService.register(user2);
+        actual = storage.get("login2");
+        assertEquals(user2,actual);
+
+        registrationService.register(user3);
+        actual = storage.get("login3");
+        assertEquals(user3,actual);
+    }
+
+    @Test
+    void getUserByUnrealLogin() {
+        assertEquals(null, storage.get("nuefdd"));
+
+        assertEquals(null, storage.get("dfghjkl"));
+    }
+
+    @Test
+    void registerAndGetUnrealUsers() {
+        User actual = storage.get("login1");
+        assertNotEquals(user2,actual);
+
+        actual = storage.get("login2");
+        assertNotEquals(user3,actual);
+
+        actual = storage.get("login3");
+        assertNotEquals(user1,actual);
+    }
+
+    @Test
+    void registerNullUser() {
+        User userNull = null;
+        assertThrows(UserIsNotValidException.class, () -> {
+            registrationService.register(userNull);
+        }, "User mustn`t be null");
+    }
+
+    @Test
+    void registerUserWithNullLogin() {
+        User userWithLogin = new User(10L, null, "password", 48);
+        assertThrows(UserIsNotValidException.class, () -> {
+            registrationService.register(userWithLogin);
+        }, "User login mustn`t be null");
+    }
+
+    @Test
+    void registerUserWithNullPassword() {
+        User userWithPassword = new User(10L, "login", null, 48);
+        assertThrows(UserIsNotValidException.class, () -> {
+            registrationService.register(userWithPassword);
+        }, "User password mustn`t be null");
+    }
+
+    @Test
+    void registerUserWithAllUnrealValues() {
+        User userWithLoginAndPasswordAndAge = new User(10L, null, null, -10);
+        assertThrows(UserIsNotValidException.class, () -> {
+            registrationService.register(userWithLoginAndPasswordAndAge);
+        }, "User have invalid value");
+    }
+
+    @Test
+    void registerUserWithNullPasswordAndLogin() {
+        User userWithLoginAndPassword = new User(10L, null, null, 30);
+        assertThrows(UserIsNotValidException.class, () -> {
+            registrationService.register(userWithLoginAndPassword);
+        }, "User login or password mustn`t be null");
+    }
+
+    @Test
+    void registerUserWithShotPassword() {
+        User userWithShortPassword = new User(10L, "login", "pass", 20);
+        assertThrows(UserIsNotValidException.class, () -> {
+            registrationService.register(userWithShortPassword);
+        }, "User password must be longer than 6 elements");
+    }
+
+    @Test
+    void registerUserWithBadAge() {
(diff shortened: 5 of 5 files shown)


src/main/java/core/basesyntax/RegistrationServiceImpl.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+import core.basesyntax.dao.StorageDao;
+import core.basesyntax.dao.StorageDaoImpl;
+//import core.basesyntax.db.Storage;
```
GalubaGSV, line 5: delete unnecessary comments

src/main/java/core/basesyntax/RegistrationServiceImpl.java
```diff
@@ -0,0 +1,28 @@
...
+        if (user == null) {
+            throw new UserIsNotValidException("user is null");
+        }
+        if (user.getLogin() == null || user.getPassword() == null) {
+            throw new UserIsNotValidException("users login or password is null");
+        }
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new UserIsNotValidException("user is already registered");
+        }
+        if (user.getAge() < 18 || user.getPassword().length() < 6) {
```
GalubaGSV, line 22: Magic numbers must be constants

src/main/java/core/basesyntax/RegistrationServiceImpl.java
```diff
@@ -0,0 +1,28 @@
...
+            throw new UserIsNotValidException("incorrect password or age");
```
GalubaGSV, line 23: add more informative message. You can add conditions to message like more 18 or more then 6 characters

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,157 @@
...
+        actual = storage.get(LOGIN3);
+        assertNotEquals(user1,actual);
+    }
+
+    @Test
+    void registerNullUser() {
+        User userNull = (User) NULL_VALUE;
+        Assertions.assertThrows(UserIsNotValidException.class, () -> {
+            registrationService.register(userNull);
+        }, "User mustn`t be null");
```
GalubaGSV, line 95: You message must be like:
"Expected " + UserIsNotValidException.class.getName() + " to be thrown for the null user, but it wasn't"
You need add to the messages what is expected
And you need remake you others messages

GalubaGSV (changes requested):
some improvements are needed

GalubaGSV (approved):
Good Job!

src/main/java/core/basesyntax/RegistrationService.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+import core.basesyntax.model.User;
+
+public interface RegistrationService {
+    User register(User user) throws UserIsNotValidException;
```
karma-o, line 6: how do we **Unchecked** Exceptions?

src/main/java/core/basesyntax/RegistrationServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+            throw new UserIsNotValidException("user is already registered");
+        }
+        if (user.getAge() < USER_AGE || user.getPassword().length() < PASSWORD_LENGTH) {
+            if (user.getAge() < USER_AGE) {
+                throw new UserIsNotValidException(
+                        "incorrect age. User should have more than 18 year old");
+            } else {
+                throw new UserIsNotValidException(
+                        "incorrect password. Password should have more than 6 elements");
+            }
```
karma-o, line 30: what does password have to do with age? Why if we have a valid age then our password is always wrong?
and I do not see any check of password length to say that it is too short.
We should have a structure like `checkPassword()`, where you check for null and length for example, and the same for other fields

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+        actual = storage.get(LOGIN3);
+        assertEquals(user3,actual);
+    }
+
+    @Test
+    void getUserByUnrealLogin() {
+        assertEquals(NULL_VALUE, storage.get(UNREAL_LOGIN1));
+
+        assertEquals(NULL_VALUE, storage.get(UNREAL_LOGIN2));
+    }
```
karma-o, line 76: the task was about writing an implementation and tests for `registrationService`, storage - is not included

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+    void registerAndGetUnrealUsers() {
+        User actual = storage.get(LOGIN1);
+        assertNotEquals(user2,actual);
+
+        actual = storage.get(LOGIN2);
+        assertNotEquals(user3,actual);
+
+        actual = storage.get(LOGIN3);
+        assertNotEquals(user1,actual);
+    }
```
karma-o, line 88: you have not called a `register()` method, so it is quite expected to return null, this case does not test anything, remove it

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+    private static RegistrationService registrationService;
+    private static StorageDao storage;
+    private User user1;
+    private User user2;
+    private User user3;
+
+    @BeforeAll
+    static void beforeAll() {
+        registrationService = new RegistrationServiceImpl();
+        storage = new StorageDaoImpl();
```
karma-o, line 46: how is this `storage` connected to the one that you have in `registrationService`?

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @BeforeEach
+    void setUp() {
+        user1 = new User(USER_ID1,LOGIN1, PASSWORD1, AGE1);
+        user2 = new User(USER_ID2,LOGIN2, PASSWORD2, AGE2);
+        user3 = new User(USER_ID3,LOGIN3, PASSWORD3, AGE3);
```
karma-o, line 53: no need for 3 users, 1 default valid user is enough, and you can just modify some of its fields up to you needs in each test case so that all other user properties would be valid and only the one that you are testing is not (in case it is `notOk` case of course)

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+    private static final long USER_ID1 = 1L;
+    private static final long USER_ID2 = 2L;
+    private static final long USER_ID3 = 3L;
+    private static RegistrationService registrationService;
+    private static StorageDao storage;
+    private User user1;
+    private User user2;
+    private User user3;
+
+    @BeforeAll
```
karma-o, line 43: add `after each` with `Storage.people.clear()`  so that your tests are independent, and data from different tests do not interfere

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+
+    private static final Object NULL_VALUE = null;
```
karma-o, line 17:
```suggestion
```

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+        User userWithLogin = new User(USER_ID1, (String) NULL_VALUE, PASSWORD, AGE2);
+        assertThrows(UserIsNotValidException.class, () -> {
+            registrationService.register(userWithLogin);
+        }, "Expected " + UserIsNotValidException.class.getName()
+                + "to be thrown for the null user login, but it wasn't");
+    }
+
+    @Test
+    void registerUserWithNullPassword() {
+        User userWithPassword = new User(USER_ID1, LOGIN, (String) NULL_VALUE, AGE1);
```
karma-o, line 110: you have set up users in the beginning, use and modify them if you need instead of creating a new one each time, same for other tests

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+    private static final String PASSWORD2 = "password2";
+    private static final String PASSWORD3 = "password3";
+    private static final int AGE1 = 22;
+    private static final int AGE2 = 40;
+    private static final int AGE3 = 18;
+    private static final int SMALL_AGE = 12;
+    private static final int UNREAL_AGE = -10;
+    private static final long USER_ID1 = 1L;
+    private static final long USER_ID2 = 2L;
+    private static final long USER_ID3 = 3L;
```
karma-o, line 36: names are not informative, is it valid data or not? existing or not?

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+    void registerUserWithShotPassword() {
+        User userWithShortPassword = new User(USER_ID1, LOGIN, SHORT_PASSWORD, AGE2);
+        assertThrows(UserIsNotValidException.class, () -> {
+            registrationService.register(userWithShortPassword);
+        }, "Expected " + UserIsNotValidException.class.getName()
+                + "to be thrown for the short user password, but it wasn't");
+    }
+
+    @Test
+    void registerUserWithBadAge() {
```
karma-o, line 146: bad -> invalid

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Test
+    void registerAndGetOrdinaryUsers() {
+        registrationService.register(user1);
+        User actual = storage.get(LOGIN1);
+        assertEquals(user1,actual);
+
+        registrationService.register(user2);
+        actual = storage.get(LOGIN2);
```
karma-o, line 63: to get the actual user, check the returning value of the method instead of the `Storage.get()`, the same for the others
\+ how can we check if the user was actually registered if we are checking in the way I described above?

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+    @Test
+    void registerTheSameUsers() {
+        assertEquals(user1, storage.get(LOGIN1));
+
+        user2 = user1;
+        assertThrows(UserIsNotValidException.class, () -> {
+            registrationService.register(user2);
+        }, "Expected " + UserIsNotValidException.class.getName()
+                + "to be thrown for the same user login, but it wasn't");
+    }
```
karma-o, line 163: can we reformat it to the test like: `register_userWithAlreadyExistingLogin_throwsException` and how should we add this existing user?

karma-o (changes requested):

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,153 @@
...
+            registrationService.register(userNull);
+        },"Expected " + UserIsNotValidException.class.getName()
+                + " to be thrown for the null user, but it wasn't");
+    }
+
+    @Test
+    void register_userWithNullLogin_isNotOk() {
+        testUser.setLogin(null);
+        testUser.setPassword(PASSWORD);
+        testUser.setAge(AGE);
```
maxline, line 67: fields PASSWORD and AGE already was set above in the method setUp(), 
it is not necessary to duplicate the values, just set values that different from the setUp() method.

Check the all others tests also.

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,153 @@
...
+        assertThrows(UserIsNotValidException.class, () -> {
+            registrationService.register(testUser);
+        }, "Expected " + UserIsNotValidException.class.getName()
+                + "to be thrown for the small user age, but it wasn't");
+    }
+
+    @Test
+    void register_theSameUsers_isNotOk() {
+        registrationService.register(testUser);
+        assertEquals(testUser, storage.get(LOGIN));
```
maxline, line 131: `actual = registrationService.register(testUser);`
instead of `storage.get()`

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,153 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static final long USER_ID = 145789L;
+    private static final String LOGIN = "Login";
+    private static final String PASSWORD = "password";
+    private static final String PASSWORD_FOR_SAME_USER = "passwordsame";
+    private static final String SHORT_PASSWORD = "pass";
+    private static final int AGE = 22;
```
maxline, line 23: I think more informative will be create constant for `MIN_AGE`, and during testing use e.g. `MIN_AGE-1` , `MIN_AGE+1`

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,153 @@
...
+    private static final int AGE_FOR_SAME_USER = 40;
```
maxline, line 24: do we need a different age for the same user or we can use the same age? 
the same question about PASSWORD_FOR_SAME_USER

maxline (changes requested):
Good job! Let's improve the solution

sarakhmen (approved):
Good job :)

PR #492, opened 2023-03-01
Solved

arthursilver111:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,33 @@
...
+        if (user.getLogin() == null) {
+            throw new CustomException("Login can not be null");
+        }
+        if (user.getPassword() == null) {
+            throw new CustomException("Password can not be null");
+        }
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new CustomException("Login exist");
+        }
+        if (user.getAge() < MIN_AGE) {
```
olekskov, line 24: are we sure on this stage that age is not null?

olekskov (changes requested):

Ivan95kos (approved):
Great job

PR #507, opened 2023-03-04
Registration validation service home work

V-Vasylchuk:


src/main/java/core/basesyntax/model/User.java
```diff
@@ -8,6 +8,13 @@ public class User {
     private String password;
     private Integer age;
 
+    public User(Long id, String login, String password, Integer age) {
+        this.id = id;
+        this.login = login;
+        this.password = password;
+        this.age = age;
+    }
```
PlagueEbola, line 16: I'm not sure that you can add a constructor to User

PlagueEbola (approved):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,60 @@
...
+    }
+
+    private void ageValidation(User user) {
+        if (user.getAge() == null) {
+            throw new RegistrationUserException("Please write your age:");
+        }
+        if (user.getAge() < MIN_AGE) {
+            throw new RegistrationUserException("You are too young!");
+        }
+        if (user.getAge() > MAX_AGE || user.getAge() < ZERO) {
```
nacenik, line 56: `user.getAge() < ZERO` this check is never used. Cause all value that are less than 0 is less than MIN_AGE too. 
```suggestion
        if (user.getAge() > MAX_AGE) {
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,60 @@
...
+            throw new RegistrationUserException("Incorrect age!");
```
nacenik, line 57: add an understandable message. Incorrect why?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,117 @@
...
+    @Test
+    void registration_withEmptyLogin_isNotOk() {
+        testUser.setLogin("");
+        assertThrows(RegistrationUserException.class,
+                () -> registrationService.register(testUser), "Login can't be empty!");
+    }
+
+    @Test
+    void registration_withSameLogin_isNotOk() {
+        User actualUser = registrationService.register(testUser);
```
nacenik, line 84: You can’t use method register() here because you testing this method. You need to add a user to the base like this: Storage.people.add().
And then register the user with the same email using register() method expecting an exception.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,117 @@
...
+    @Test
+    void registration_withGreaterAge_isNotOk() {
+        testUser.setAge(GREATER_AGE);
+        assertThrows(RegistrationUserException.class,
+                () -> registrationService.register(testUser), "Age can't be greater than 125");
+    }
+
+    @Test
+    void registration_withNegativeAge_isNotOk() {
+        testUser.setAge(ZERO);
```
nacenik, line 113: 0 in not negative.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,117 @@
...
+    @Test
+    void registration_withNullAge_isNotOk() {
+        testUser.setAge(null);
+        assertThrows(RegistrationUserException.class,
+                () -> registrationService.register(testUser), "Age can't be null!");
+    }
+
+    @Test
+    void registration_withLessThan18YearsOld_isNotOk() {
+        testUser.setAge(LESS_AGE);
```
nacenik, line 99: for testing it is better to take marginally unacceptable input data, in this case 17 years

nacenik (changes requested):
Let’s improve your solution ;)

PR #524, opened 2023-03-05
Maksym Yashyn HomeWork

Nimakel:


NickolayVoloshchuk (approved):
I made some comments, but I don't know if they are significant, so I approved. But with the names of the methods, you better fix it, in the readme file of the project it is written how the test methods should be called specifically.

src/main/java/core/basesyntax/service/RegistrationFailedException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+public class RegistrationFailedException extends Exception {
```
karma-o, line 3:
```suggestion
public class RegistrationFailedException extends RuntimeException {
```

src/main/java/core/basesyntax/service/RegistrationService.java
```diff
@@ -3,5 +3,5 @@
 import core.basesyntax.model.User;
 
 public interface RegistrationService {
-    User register(User user);
+    User register(User user) throws RegistrationFailedException;
```
karma-o, line 6:
```suggestion
    User register(User user);
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,23 @@
...
+    private static final int MIN_PASS_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
-    public User register(User user) {
-        return null;
+    public User register(User user) throws RegistrationFailedException {
+        if (user.getAge() == null
+                || user.getAge() < MIN_AGE) {
+            throw new RegistrationFailedException("Your age must be over 18!");
```
karma-o, line 16:
```suggestion
            throw new RegistrationFailedException("Your age must be over " + MIN_AGE);
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,23 @@
...
+        } else if (user.getPassword() == null
+                || user.getPassword().length() <= MIN_PASS_LENGTH) {
+            throw new RegistrationFailedException("Your password must be at least 6 characters!");
+        } else if (user.getLogin() == null) {
+            throw new RegistrationFailedException("You wrote an empty login!");
+        } else if (storageDao.get(user.getLogin()) != null) {
+            throw new RegistrationFailedException("Your login already exist!");
+        }
```
karma-o, line 24: do not overcomplicate `if else`s, just have several `if`s one by one without else, because there is no real reason to add them

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+        userThree.setAge(18);
+
+        userFour.setLogin("ILoveCherry");
+        userFour.setPassword("ireallylovecherry");
+        userFour.setAge(21);
+
+        userFive.setLogin(null);
+        userFive.setPassword(null);
+        userFive.setAge(null);
+    }
```
karma-o, line 39: * 1 valid user is enough, you just need to clear DB after each test to reuse it
* initialize it with a `new User()` before each test to avoid having common data for several tests.
* set user some 1 corresponding non-valid field in each `notOk` test case
* `Before All` itself should only initialize `registrationService`
* use constants for user data, so that it would be clear what is valid and what is not

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static User userOne = new User();
+    private static User userTwo = new User();
+    private static User userThree = new User();
+    private static User userFour = new User();
+    private static User userFive = new User();
+    private RegistrationServiceImpl registrationService = new RegistrationServiceImpl();
```
karma-o, line 16: Use abstraction
```suggestion
    private RegistrationService registrationService;
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+    void userAge_Ok() {
+        try {
+            assertEquals(userFour, registrationService.register(userFour));
+        } catch (RegistrationFailedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Test
+    void userAge_NotOk() {
```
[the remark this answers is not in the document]
karma-o: be attentive, you missed a 3rd part of the method name

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+        userFour.setAge(21);
+
+        userFive.setLogin(null);
+        userFive.setPassword(null);
+        userFive.setAge(null);
+    }
+
+    @Test
+    void userAge_Ok() {
+        try {
```
karma-o, line 43: remove all try\catch blocks

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+        } catch (RegistrationFailedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Test
+    void userAge_NotOk() {
+        assertThrows(RegistrationFailedException.class, () -> {
+            registrationService.register(userOne);
+        }, "Your age must be over 18!");
```
karma-o, line 54: assertion messages are not the same as the exception messages, first ones are telling us the reason why the test failed, in this case, it should be 
```
" expected " + RegistrationFailedException.class.getName() + " to be thrown for age " + user.getAge() + " but it wasn't"
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Test
+    void userLogin_Ok() {
+        try {
+            assertEquals(userFour, registrationService.register(userFour));
+        } catch (RegistrationFailedException e) {
+            throw new RuntimeException(e);
+        }
+    }
```
karma-o, line 67: all of your `ok` cases are exact duplicates, leave only 1 of them, name it `register_validUser_ok()` and check the id assigning in there

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Test
+    void userLogin_NotOk() {
+        assertThrows(RegistrationFailedException.class, () -> {
+            registrationService.register(userThree);
+        }, "Your login already exist!");
+        assertThrows(RegistrationFailedException.class, () -> {
+            registrationService.register(userFive);
+        }, "You wrote an empty login!");
```
karma-o, line 76: I can see an attempt to write the existing user test case, we should add it via `Storage.people.add(user)`, and then try to register it

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+    void userPassword_Ok() {
+        try {
+            assertEquals(userFour, registrationService.register(userFour));
+        } catch (RegistrationFailedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Test
+    void userPassword_NotOk() {
```
karma-o, line 89: overall you have too few test cases, every user field(except for id, since it is not registered yet) should have a null check, and some business requirement checks like non-valid password length or invalid age cases, etc.

karma-o (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,167 @@
...
+        }, " expected " + RegistrationFailedException.class.getName()
+                + " to be thrown for over age "
+                + user.getAge()
+                + " but it wasn't");
+    }
+
+    @Test
+    void register_successful_ok() {
+        registrationService.register(user);
+        Storage.people.add(user);
```
nacenik, line 164: this is useless line, cause `registrationService.register(user);` have already added new user

nacenik (approved):
LGTM! 🚀

PR #555, opened 2023-03-23
solved task

slaybrute:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -9,6 +9,24 @@ public class RegistrationServiceImpl implements RegistrationService {
 
     @Override
     public User register(User user) {
-        return null;
+        if (user.getLogin() == null) {
```
kozhukhovsky, line 12: what happens if the user is null?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -9,6 +9,24 @@ public class RegistrationServiceImpl implements RegistrationService {
...
+        if (user.getPassword() == null) {
+            throw new RegistrationException("Password can't be null");
+        }
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new RegistrationException("Such login is already existed");
+        }
+        if (user.getAge() < 18) {
+            throw new RegistrationException("User should be 18 and older");
+        }
+        if (user.getPassword().toCharArray().length < 6) {
```
kozhukhovsky, line 27:
```suggestion
        if (user.getPassword().length() < 6) {
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+        userWithNullPassword = new User();
+        userWithNullPassword.setLogin("user9");
+        userWithNullPassword.setPassword(null);
+        userWithNullPassword.setAge(28);
+
+    }
+
+    @Test
+    @Order(1)
```
kozhukhovsky, line 93: It is better not to use the order of tests, they should not depend one on one.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,161 @@
...
+        assertEquals(2, userList.size());
+    }
+
+    @Test
+    @Order(6)
+    void register_validData2_Ok() {
+        User user = registrationService.register(userWithValidData2);
+        assertEquals(userWithValidData2, user);
+        assertEquals(3, userList.size());
+    }
```
kozhukhovsky, line 135:
```suggestion
    @Test
    @Order(5)
    void register_validData1_Ok() {
        User user = registrationService.register(userWithValidData1);
        assertEquals(userWithValidData1, user);
        assertEquals(2, userList.size());
    }

    @Test
    @Order(6)
    void register_validData2_Ok() {
        User user = registrationService.register(userWithValidData2);
        assertEquals(userWithValidData2, user);
        assertEquals(3, userList.size());
    }
```

No need to do the same tests for the same functionality. If you want to test multiple users, just add it to one test.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @Test
+    @Order(7)
+    void register_sameLogin2_notOk() {
```
kozhukhovsky, line 139: same

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,161 @@
...
+    private static User userWithNullPassword;
+    private RegistrationServiceImpl registrationService = new RegistrationServiceImpl();
+
+    @AfterEach
+    void tearDown() {
+        userList = Storage.people;
+    }
+
+    @BeforeAll
+    static void beforeAll() {
```
kozhukhovsky, line 37: Does not need to be used before all user tests for all tests. What if you run one test?
Just create a user, and make the conditions you need in each test.

kozhukhovsky (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,148 @@
...
+        userWithInvalidAge = new User();
+        userWithInvalidAge.setPassword("1234556");
+        userWithInvalidAge.setLogin("user1");
+        userWithInvalidAge.setAge(10);
+        assertThrows(RegistrationException.class, () ->
+                registrationService.register(userWithInvalidAge));
+    }
+
+    @Test
+    @Order(1)
```
uliana-andrushchuk, line 53: Please, remove `Order` annotation from all tests

uliana-andrushchuk (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.User;
+import java.util.List;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static List<User> userList;
```
okuzan, line 14: useless variable

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+    private RegistrationServiceImpl registrationService = new RegistrationServiceImpl();
+
+    @AfterEach
+    void tearDown() {
+        userList = Storage.people;
+    }
+
+    @BeforeAll
+    static void beforeAll() {
+        userList = Storage.people;
```
okuzan, line 35: better move registration service initialization here

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+    private static User userWithInvalidAge;
+    private static User userWithValidData;
+    private static User userWithValidDateSameLogin;
+    private static User userWithInvalidPassword;
+    private static User userWithValidData1;
+    private static User userWithValidData2;
+    private static User userWithValidDateSameLogin2;
+    private static User userWithNullAge;
+    private static User userWithNullLogin;
+    private static User userWithNullPassword;
```
okuzan, line 25: numbers in variable naming is a bad practice. I suggest cleaning this up and leaving only one user, which you will set up with valid attributes in BeforeEach, and then amend to your needs in each test like this:
image

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+        userWithInvalidAge = new User();
+        userWithInvalidAge.setPassword("1234556");
+        userWithInvalidAge.setLogin("user1");
+        userWithInvalidAge.setAge(10);
+        assertThrows(RegistrationException.class, () ->
+                registrationService.register(userWithInvalidAge));
+    }
+
+    @Test
+    void register1_validData_Ok() {
```
okuzan, line 49: testing only one user here is enough

okuzan (changes requested):
Would be nice to check exception messages too, for example, if null user fails on null password check, you would immediately know that although the invalid user isn't able to register, your tests suite doesn't work as expected

uliana-andrushchuk (approved):

PR #556, opened 2023-03-24
registration-validation implementation

Dmytro-Byrka:


Dmytro-Byrka:
Hi, have one problem with coverage, expected coverage should be 90+ but i have 86 and honestly dont know how make it bigger

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+    @Test
+    void register_UserNegativeAge_NotOk() {
+        user.setAge(-25);
+        assertThrows(RegistrationException.class, () -> {
+            registrationService.register(user);
+        });
+    }
+
+    @Test
+    void register_TwoDifferentUsers_Ok() {
```
okuzan, line 142: I'm not sure what this checks and whether we need it

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,164 @@
...
+    static void beforeAll() {
+        user = new User();
+        anotherUser = new User();
+        registrationService = new RegistrationServiceImpl();
+        storageDao = new StorageDaoImpl();
+    }
+
+    @BeforeEach
+    void setUp() {
+        anotherUser.setId(44L);
```
okuzan, line 32: we don't need to set ids preemptively for test users

okuzan (changes requested):
would be nice to test exception messages too

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Dmytro-Byrka: I was trying to test normal cases, when we add to different users in storage and checking that they are really different

Dmytro-Byrka: Interesting, when i removed that test case, test coverage icreased to 95%

Dmytro-Byrka:
> would be nice to test exception messages too

Screenshot_1

Can i do exception messages testing like this? Only one problem, i hardcoded some values, that are private constants in RegistrationServiceImpl class

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Dmytro-Byrka: Do you meen for every test case or overrall we dont need to test it?

PR #561, opened 2023-04-04
Task is registration-validation. I did 6 test methods, but i don't kn…

Andrei-Bondarenko-MRPL:
…ow, how should i make the comparison so that the age is more than 18.


Rommelua (changes requested):
Java Cl checks failed. Before making a Pull Request (or a new commit to an already existing PR), you need to write the mvn clean package command in the terminal and make sure that all checks pass. After that, the mentor will be able to review the homework

image

PR #564, opened 2023-04-08
implemented test for RegistrationServiceImol class

0-zagar-0:


src/main/java/core/basesyntax/exception/ValidationException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.exception;
+
+public class ValidationException extends Exception {
```
andriy-shulepa, line 3: Consider extending RuntimeException instead of Exception, as it does not require the calling method to handle the exception explicitly.

src/main/java/core/basesyntax/model/User.java
```diff
@@ -1,61 +1,68 @@
...
+
+import java.util.Objects;
+
+public class User {
+    private Long id;
+    private String login;
+    private String password;
+    private Integer age;
+    
+    public User(Long id, String login, String password, Integer age) {
```
andriy-shulepa, line 11: Consider using a constructor without the id parameter, as the id is usually generated when a user is persisted to a database. Creating a constructor without the id parameter would make it easier for users to create a new User object.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,33 @@
...
+                || user.getPassword() == null
+                || user.getAge() == null) {
+            throw new ValidationException("Any user value cannot be null");
+        }
+        
+        if (user.getId().toString().length() >= MIN_ID_LENGTH
+                && storageDao.get(user.getLogin()) == null
+                && user.getLogin().length() >= MIN_LOGIN_AND_PASSWORD_LENGTH
+                && user.getPassword().length() >= MIN_LOGIN_AND_PASSWORD_LENGTH
+                && user.getAge() >= MIN_USER_AGE) {
```
andriy-shulepa, line 27: Instead of having all conditions in one if statement, consider breaking it into separate if statements with specific ValidationException messages for each condition. This will help users understand what went wrong during registration.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,33 @@
...
+            storageDao.add(user);
+        } else {
+            return user;
```
andriy-shulepa, line 30: Returning the same user object when registration validation fails can be confusing. Consider throwing a ValidationException with a specific message to indicate the reason for the failure.

src/main/java/core/basesyntax/service/RegistrationService.java
```diff
@@ -1,7 +1,8 @@
 package core.basesyntax.service;
 
+import core.basesyntax.exception.ValidationException;
 import core.basesyntax.model.User;
 
 public interface RegistrationService {
-    User register(User user);
+    User register(User user) throws ValidationException;
```
andriy-shulepa, line 7:
```suggestion
    User register(User user);
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,33 @@
...
+    private static final int MIN_LOGIN_AND_PASSWORD_LENGTH = 6;
+    private static final int MIN_USER_AGE = 18;
+    private static final int MIN_ID_LENGTH = 8;
     private final StorageDao storageDao = new StorageDaoImpl();
-
+    
     @Override
-    public User register(User user) {
+    public User register(User user) throws ValidationException {
+        if (user.getId() == null
```
andriy-shulepa, line 16: The id should be generating during persisting user, no need to validate it here

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,33 @@
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exception.ValidationException;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_LOGIN_AND_PASSWORD_LENGTH = 6;
```
andriy-shulepa, line 9: you should have two separate constants even if they have the same value. constraints for login and password length can change independently

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,75 @@
...
+    @BeforeEach
+    void setUp() {
+        Storage.people.add(new User(123456789L, "Bob123", "111111111", 19));
+        Storage.people.add(new User(123456789L, "Alice123", "222222222", 22));
+        Storage.people.add(new User(123456789L, "John123", "333333333", 25));
+        size = Storage.people.size();
+    }
+    
+    @Test
+    void notContainsUserInStorage_Ok() throws ValidationException {
```
andriy-shulepa, line 27: please, follow naming convention from the task in all tests
image

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,33 @@
...
+                || user.getAge() == null) {
+            throw new ValidationException("Any user value cannot be null");
+        }
+        
+        if (user.getId().toString().length() >= MIN_ID_LENGTH
+                && storageDao.get(user.getLogin()) == null
+                && user.getLogin().length() >= MIN_LOGIN_AND_PASSWORD_LENGTH
+                && user.getPassword().length() >= MIN_LOGIN_AND_PASSWORD_LENGTH
+                && user.getAge() >= MIN_USER_AGE) {
+            storageDao.add(user);
```
andriy-shulepa, line 28: you method should return saved user instance if it was saved successfully and throw validation exception in other case

andriy-shulepa (changes requested):
The test's logic wasn't reviewed since the main logic should be changed. But you don't have to assert storage but assert expected user to be returned and expected error to be thrown

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+    private final RegistrationService registrationService = new RegistrationServiceImpl();
+    private int size;
+    
+    @BeforeEach
+    void setUp() {
+        Storage.people.add(new User("Bob123", "111111111", 19));
+        Storage.people.add(new User("Alice123", "222222222", 22));
+        Storage.people.add(new User("John123", "333333333", 25));
+        size = Storage.people.size();
+    }
```
Ivan95kos, line 23: useless users, you do not use them in tests
```suggestion
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+    
+    @Test
+    void register_nonExistentLogin_Ok() {
+        User actual = registrationService.register(new User("Kate123", "111111111", 19));
+        assertEquals(size + 1, Storage.people.size());
+        assertNotNull(actual);
+    }
```
Ivan95kos, line 30: better to use StorageDao interface
```suggestion
    @Test
    void register_successful() {
        User actual = registrationService.register(new User("Kate123", "111111111", 19));
        assertEquals(actual, storageDao.get(actual.getLogin()));
    }

```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+    
+    @Test
+    void register_anExistingLogin_NotOk() {
+        assertThrows(ValidationException.class,() -> {
+            registrationService.register(new User("Bob123", "111111111", 19));
+        });
+        assertEquals(size, Storage.people.size());
```
Ivan95kos, line 37: useless check, we have assertThrows, that is enough
```suggestion
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+    
+    @Test
+    void register_shortLogin_NotOk() {
+        assertThrows(ValidationException.class, () -> {
+            registrationService.register(new User("Bob", "165498432156", 19));
+        });
+        assertEquals(size, Storage.people.size());
```
Ivan95kos, line 45: the same
```suggestion
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+    
+    @Test
+    void register_shortPassword_NotOk() {
+        assertThrows(ValidationException.class, () -> {
+            registrationService.register(new User("Clark123", "11", 19));
+        });
+        assertEquals(size, Storage.people.size());
```
Ivan95kos, line 53:
```suggestion
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+    
+    @Test
+    void register_lessThan18YearsOld_NotOk() {
+        assertThrows(ValidationException.class, () -> {
+            registrationService.register(new User("Ivan123", "419813541", 17));
+        });
+        assertEquals(size, Storage.people.size());
```
Ivan95kos, line 61:
```suggestion
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,36 @@
...
+        if (user.getLogin().length() < MIN_LOGIN_LENGTH) {
+            throw new ValidationException("Login is too short, it must be more than 6 characters");
+        }
+        if (user.getPassword().length() < MIN_PASSWORD_LENGTH) {
+            throw new ValidationException(
+                    "Password is too short, it must be more than 6 characters");
+        }
+        if (user.getAge() < MIN_USER_AGE) {
+            throw new ValidationException("User must be over 18 years of age");
+        }
```
Ivan95kos, line 34: Hard to read.
It is better to divide into methods and describe the necessary logic in them.
```
private checkAge(User user) {
// age verification
}
```

Ivan95kos (changes requested):

src/main/java/core/basesyntax/exception/ValidationException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.exception;
+
+public class ValidationException extends RuntimeException {
```
Ivan95kos, line 3: For the future, it is better to give more specific name
```suggestion
public class RegistrationException extends RuntimeException {
```

Ivan95kos (approved):
Good job overall!

PR #581, opened 2023-05-03
solution

VictoriaMrkvch:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,36 @@
...
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_AGE = 18;
+    private static final int MIN_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user.getLogin() == null) {
```
kozhukhovsky, line 15: What happens if the user is null?

kozhukhovsky (changes requested):

olekskov (approved):
Good job

PR #584, opened 2023-05-07
register() implemented, tests provided.

TymurProkhorov:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,29 @@
...
+import core.basesyntax.db.Storage;
+import core.basesyntax.exception.ValidationException;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
```
uliana-andrushchuk, line 12: Please, add check if user is null, if user.getLogin() is null and user.getPassword() is null

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,29 @@
...
+        for (User eachUser : Storage.people) {
+            if (eachUser.getLogin().equals(user.getLogin())) {
```
uliana-andrushchuk, line 15: You can simply check if the user exists using this `storageDao.get(login) != null`

uliana-andrushchuk (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+    void setUp() {
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @Test
+    void containsValidUser_Ok() {
+        User kate = new User();
+        kate.setAge(40);
+        kate.setLogin("1234566666");
+        kate.setPassword("43434343");
```
olekskov, line 25: Please move creation of default ok user to BeforeEach method and in tests only change 1 required field(e.g. set login to null, or password to short one, etc) and perform test.
This will significantly reduce amount of code and highlight what actually we test

olekskov (changes requested):

olekskov (approved):
Good job

PR #589, opened 2023-05-11
task solution implemented

Tupadzaky:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static User user;
+    private static RegistrationService registrationService;
+    private static final long ID = 0;
+    private static final String LOGIN_VALID = "aboba@gmail.com";
+    private static final String PASSWORD_VALID = "123456";
+    private static final String PASSWORD_AND_EMAIL_INVALID = "12345";
```
sZlishchev, line 18: In my opinion it's better to create invalid field for email and password separately

sZlishchev (changes requested):
Good job! Let's add one minor improvement

src/main/java/core/basesyntax/middleware/user/stringparams/login/LoginExistsMiddleware.java
```diff
@@ -0,0 +1,18 @@
...
+import core.basesyntax.model.User;
+
+public class LoginExistsMiddleware extends Middleware {
+    @Override
+    public void check(User user) {
+        for (User inStorage : Storage.people) {
+            if (user.getLogin().equals(inStorage.getLogin())) {
+                throw new RegistrationException("User with such email already exists");
+            }
+        }
```
sarakhmen, line 15: Can we use storageDao.get() method here instead?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+class RegistrationServiceImplTest {
+    private static User user;
+    private static RegistrationService registrationService;
+    private static final long ID = 0;
+    private static final String LOGIN_VALID = "aboba@gmail.com";
+    private static final String PASSWORD_VALID = "123456";
+    private static final String LOGIN_INVALID = "aboba";
+    private static final String PASSWORD_INVALID = "12345";
+
```
sarakhmen, line 20:
```suggestion
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,100 @@
...
+        user.setAge(AGE_VALID);
+        user.setLogin(LOGIN_VALID);
+        user.setPassword(PASSWORD_VALID);
+    }
+
+    @Test
+    void register_nullUser_notOk() {
+        assertThrows(RegistrationException.class,
+                () -> registrationService.register(null),
+                "User shouldn't be null");
```
sarakhmen, line 42: If I'm not mistaken this 3-rd parameter (message) is used when an exception is not raised. Here are some docs:
image

So, it's better to provide something like 'RegistrationException error was expected' instead

sarakhmen (changes requested):
Interesting solution, good job ;) Let's only slightly improve it

sarakhmen (approved):
LGTM! :)

PR #591, opened 2023-05-16
implemented task

chrisnyrka:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,127 @@
...
+    private static final String EMPTY_LOGIN = "";
+    private static final int VALID_AGE = 22;
+    private static final int ENOUGH_AGE = 18;
+    private static final int NOT_ENOUGH_AGE = 12;
+    private static final int NEGATIVE_AGE = -10;
+    private static final String LONG_PASSWORD = "robinson134";
+    private static final String SHORT_PASSWORD = "984";
+    private static final String GOOD_PASSWORD = "123456";
+    private static final String PASSWORD_WITH_SPACE = "robi   ";
+    private static final String EMPTY_PASSWORD = "";
```
marimarig, line 26: Maybe it`s better not to add so many static final fields. Just my opinion

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,127 @@
...
+        registrationService = new RegistrationServiceImpl();
+        storageDao = new StorageDaoImpl();
+    }
+
+    @BeforeEach
+    void setUp() {
+        Storage.people.clear();
+        firstUser = new User(GOOD_LOGIN, LONG_PASSWORD, VALID_AGE);
+        secondUser = new User(LONG_LOGIN, GOOD_PASSWORD, ENOUGH_AGE);
+        thirdUser = new User(GOOD_LOGIN, GOOD_PASSWORD, VALID_AGE);
```
marimarig, line 44: I think it is not necessary to check it on 3 users for this task, one is enough, but let the mentor say how it is better.

marimarig (approved):
that's fine, but I have a few suggestions

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,44 @@
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
-    private final StorageDao storageDao = new StorageDaoImpl();
+    private static final int MIN_AGE = 18;
+    private static final int MIN_PASSWORD_AND_LOGIN_LENGTH = 6;
```
rovein, line 9: Let's separate them: `MIN_PASSWORD_SIZE` and `MN_LOGIN_SIZE`. The requirements for min length can change in the future, so we need to separate them.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,44 @@
...
+        return checkedUserForRegistration(user);
+    }
+
+    private User validUser(User user) throws RegistrationException {
+        if (user == null) {
+            throw new RegistrationException("User can't be null");
+        }
+        if (user.getLogin() == null) {
+            throw new RegistrationException("Login can't be null");
+        }
```
rovein, line 28: Let's create three different methods: `validateLogin`, `validatePassword`, `validateAge`. 

Keep there both logic for null/empty checks and for boundaries. Call these three methods in your `register`. Then return the user from dao.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
rovein: In your test classes that is OK if you use literals directly instead of constants. But, if you already have created a lot of constants, you can move them to a separate class, e.g. `RegistrationServiceTestUtils` or `RegistrationServiceTestConstants`.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
rovein: It would be simpler to not store users in class fields. Better create them directly in your test methods (as many as you have to: one, two, three, considering the test case), and fill with the needed data.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    @Test
+    void userIsAlreadyRegistered() {
```
rovein, line 48: Remember the naming convention in this task.

<methodUnderTest>_<state>_<expectedBehavior>

Example: `register_nullAge_notOk`

Fix in all other places.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,44 @@
...
+
+    private User checkedUserForRegistration(User user) {
+        if (storageDao.get(user.getLogin()) == null) {
+            if (user.getLogin().trim().length() >= MIN_PASSWORD_AND_LOGIN_LENGTH
+                    && user.getPassword().trim().length() >= MIN_PASSWORD_AND_LOGIN_LENGTH
+                    && user.getAge() >= MIN_AGE) {
+                return storageDao.add(user);
+            }
+        }
         return null;
```
rovein, line 46: You should not return the `null`. If validation fails: you need to throw your exception. See the previous comment.

Considering it, you will need to update your tests to use `assertThrows` in the places where you used `assertEquals`.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,127 @@
...
+        storageDao.add(firstUser);
+        User actual = registrationService.register(firstUser);
+        Assertions.assertNotEquals(firstUser, actual);
```
rovein, line 51: Better use `assertThrows.` See previous comments. Change your logic in `RegistrationServiceImpl`.

Or, better try TDD approach (test driven development) - firstly change the test, then change the code you are testing until the test passed.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,127 @@
...
+        User actualFirstUser = registrationService.register(firstUser);
+        User actualSecondUser = registrationService.register(secondUser);
+        User actualThirdUser = registrationService.register(thirdUser);
+        Assertions.assertNotEquals(firstUser, actualFirstUser);
+        Assertions.assertNotEquals(secondUser, actualSecondUser);
+        Assertions.assertNotEquals(thirdUser, actualThirdUser);
+    }
+
+    @Test
+    void invalidAge_NotOk() {
```
rovein, line 88: Let's split it on the several methods covering different cases, for example:

- register_minAgeUser_Ok
- register_overMinAgeUser_Ok
- register_underMinAgeUser_notOk
- register_negativeAge_notOk
- register_nullAge_notOk

Apply that approach to all other checks.

rovein (changes requested):
Good job! But let's improve it :)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
sarakhmen: Please follow @rovein advice. You can also move null-user check to a separate method. You can move logic related to user existence check to your 'validateLogin' method. Also let's return void in your validation-related methods

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,192 @@
...
+    void setUp() {
+        Storage.people.clear();
+        user = new User(RegistrationServiceTestConstants.OVER_MIN_LOGIN,
+                RegistrationServiceTestConstants.OVER_MIN_PASSWORD,
+                RegistrationServiceTestConstants.OVER_MIN_AGE);
+    }
+
+    @Test
+    void userIsAlreadyRegistered_NotOk() {
+        storageDao.add(user);
```
sarakhmen, line 33: as described in common mistakes, let's use Storage.people.add() instead

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,192 @@
...
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static RegistrationService registrationService;
+    private static StorageDao storageDao;
```
sarakhmen, line 14:
```suggestion
```
this field is unnecessary

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,192 @@
...
+    @BeforeEach
+    void setUp() {
+        Storage.people.clear();
+        user = new User(RegistrationServiceTestConstants.OVER_MIN_LOGIN,
+                RegistrationServiceTestConstants.OVER_MIN_PASSWORD,
+                RegistrationServiceTestConstants.OVER_MIN_AGE);
+    }
+
+    @Test
+    void userIsAlreadyRegistered_NotOk() {
```
sarakhmen, line 32: please pay attention to your test methods namings. Let's follow the next naming convention : `<methodUnderTest>_<state>_<expectedBehavior>` Fix the problem in all places

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,192 @@
...
+        private static final int OVER_MIN_AGE = 22;
+        private static final int MIN_AGE = 18;
+        private static final int NOT_ENOUGH_AGE = 12;
+        private static final int NEGATIVE_AGE = -10;
+        private static final String OVER_MIN_PASSWORD = "robinson134";
+        private static final String SHORT_PASSWORD = "984";
+        private static final String MIN_PASSWORD = "123456";
+        private static final String PASSWORD_WITH_SPACE = "robi   ";
+        private static final String EMPTY_PASSWORD = "";
+    }
```
sarakhmen, line 191: I think @rovein meant that this class should be a regular one, not an inner class

sarakhmen (changes requested):
Let's improve your solution one more time ;)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,61 @@
...
+    public RegistrationServiceImpl() {
+        storageDao = new StorageDaoImpl();
+    }
 
     @Override
     public User register(User user) {
-        return null;
+        validateUser(user);
+        checkUserExistence(user);
+        validateLogin(user);
```
sarakhmen, line 21: I'd recommend swapping the calls of these methods

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,171 @@
...
+    @Test
+    void register_existingUser_NotOk() {
+        Storage.people.add(user);
+        Assertions.assertThrows(RegistrationException.class, () -> {
+            registrationService.register(user);
+        });
+    }
+
+    @Test
+    void register_userNull_NotOk() {
```
sarakhmen, line 36:
```suggestion
    void register_userNull_notOk() {
```
please fix the same issue in all places

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,171 @@
...
+    @Test
+    void register_userWithSpaceLogin_NotOk() {
+        user.setLogin(RegistrationServiceTestConstants.LOGIN_WITH_SPACE);
+        Assertions.assertThrows(RegistrationException.class, () -> {
+            registrationService.register(user);
+        });
+    }
+
+    @Test
+    void register_userMinLogin_Ok() {
```
sarakhmen, line 75:
```suggestion
    void register_userMinLogin_ok() {
```

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #609, opened 2023-05-18
done the Registration Service task

ViktoriiaSemenchenko:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,64 @@
...
 
     @Override
     public User register(User user) {
-        return null;
+        isUserNull(user);
+        isLoginValid(user);
+        isPasswordValid(user);
+        isAgeValid(user);
+        return storageDao.add(user);
     }
```
nechupara, line 21: It's just recommendation to consider - maybe for better readability it's better to move all the methods with validation logic to one method (for example validateUser(user)) and place all these method's call in it
```suggestion
        validateUser(user);
        return storageDao.add(user);
    }
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,64 @@
...
+            throw new InvalidDataException("Password must be longer than "
+                    + PASSWORD_LENGTH + " characters!");
+        }
+    }
+
+    private void isAgeValid(User user) {
+        if (user.getAge() == null) {
+            throw new InvalidDataException("Age shouldn't be null!");
+        }
+        if (user.getAge() < MIN_AGE) {
```
nechupara, line 60: I don't know why, but just to mention that in the article common mistakes it is said that we should also check for negative age (but I don't understand why, so it's up to you)

nechupara (changes requested):
Good job! Let's just consider my suggestion and reject them )))

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
ViktoriiaSemenchenko: I considered such a decision and I think let the mentors tell me what will be better, and I will gladly correct it if necessary

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
ViktoriiaSemenchenko: I think that my age checks are enough

nechupara (approved):
good job )

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Test
+    void register_minAge_ok() {
+        user.setAge(MIN_AGE);
+        User registeredUser = registrationService.register(user);
+        assertSame(storageDao.get(DEFAULT_LOGIN), registeredUser);
+    }
+
+    @Test
+    public void register_nullLogin_notOk() {
```
maxline, line 57: why all tests don't have public modifiers except this one?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,64 @@
...
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_AGE = 18;
+    private static final int PASSWORD_LENGTH = 6;
+    private static final int LOGIN_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        isUserNull(user);
```
maxline, line 16: for these methods it is better to use validateUser() validateLogin() etc name style, because its return nothing.
`is` prefix usually used for the methods that returns boolean.

maxline (approved):
Well done! Approved with couple minor suggestions. Please take a look for the future.

PR #613, opened 2023-05-18
unit test written successfully

m-pnsk:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @BeforeEach
+    void setUp() {
+        user = new User();
+        user.setLogin("validlogin@gmail.com");
+        user.setPassword("valid_password");
+        user.setAge(18);
+    }
```
Valdemar619, line 29: for me it's better to have constant in the start of the class

Valdemar619 (approved):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
m-pnsk: In my opinion, this is unnecessary in unit tests

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,71 @@
...
+        validatePassword(user);
+        validateAge(user);
+        verifyUserExist(user);
+        return storageDao.add(user);
+    }
+
+    private void verifyUserForNull(User user) {
+        if (user == null) {
+            throw new NonValidUserDataException(NonValidUserDataException.class.getName()
+                    + "User is null");
```
maxline, line 27: Why do we need  `NonValidUserDataException.class.getName() info` ? This will be present in stack trace.
```suggestion
            throw new NonValidUserDataException("User is null");
```
Fix all places in the code please

src/main/java/core/basesyntax/service/NonValidUserDataException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+public class NonValidUserDataException extends RuntimeException {
```
maxline, line 3:
```suggestion
public class InvalidUserDataException extends RuntimeException {

```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Test
+    void register_userAlreadyExist_notOk() {
+        register_validUser_ok();
+        assertThrows(NonValidUserDataException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void register_getCorrectIdOfUser_Ok() {
+        register_validUser_ok();
```
maxline, line 79: do not call unit test method from the other one. you can call additional private methods if needed, but not other unit test marked by @Test annotation

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+        User newUser = new User();
+        newUser.setLogin("second_valid_login@ukr.net");
+        newUser.setPassword("password");
+        newUser.setAge(25);
+        long actualId = registrationService.register(newUser).getId();
+        long expectedId = 2L;
+        assertEquals(expectedId, actualId, "Expected ID: " + expectedId + "but was: " + actualId);
```
maxline, line 86: it this test instead of testing the id, you should check that user successfully added to the store. See the example:
```
assertTrue(Storage.people.contains(user), "User with valid password should be registered");
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @AfterEach
+    void clear() {
+        Storage.people.clear();
+        StorageDaoImpl.clearIndex();
```
maxline, line 92: I think clearIndex() is redundant operation, because in test we shouldn't assert ids, we never don't know which ids will be present in db in real application.

maxline (changes requested):
Good job! let's improve the solution

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
m-pnsk: No. This test does not check if the user is added to the list. This test checks if the user is assigned the correct index

src/main/java/core/basesyntax/model/User.java
```diff
@@ -58,4 +58,13 @@ public boolean equals(Object o) {
...
     }
+
+    @Override
+    public User clone() {
+        try {
+            return (User) super.clone();
+        } catch (CloneNotSupportedException e) {
+            throw new RuntimeException(e);
+        }
+    }
```
sarakhmen, line 69: let's revert these changes. Actually it's redundant and you don't really need to clone your user

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+        user = new User();
+        user.setLogin("validlogin@gmail.com");
+        user.setPassword("valid_password");
+        user.setAge(18);
+    }
+
+    @Test
+    void register_validUser_ok() {
+        User newUser = user.clone();
+        assertEquals(user, registrationService.register(newUser), "Users should be equal.");
```
sarakhmen, line 34: I think it's better to check if Storage.people contains our user after registration

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+    void register_userWithAgeLessThan18_notOk() {
+        user.setAge(17);
+        assertThrows(InvalidUserDataException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void register_userWithNullAge_notOk() {
+        user.setAge(null);
+        assertThrows(InvalidUserDataException.class, () -> registrationService.register(user));
+    }
```
sarakhmen, line 70: previously you grouped test cases for login validation. So, let's make your code consistent and either group age-related test cases or spilt login/password validation test case into several ones

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Test
+    void register_userAlreadyExist_notOk() {
+        registrationService.register(user);
```
sarakhmen, line 74: as described in common mistakes, let's use Storage.people.add() here

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+    @Test
+    void register_checkingCorrectnessOfUserNumbering_Ok() {
+        register_validUser_ok();
+        User newUser = new User();
+        newUser.setLogin("second_valid_login@ukr.net");
+        newUser.setPassword("password");
+        newUser.setAge(25);
+        long actualId = registrationService.register(newUser).getId();
+        long expectedId = 2L;
+        assertEquals(expectedId, actualId, "Expected ID: " + expectedId + "but was: " + actualId);
```
sarakhmen, line 87: redundant test case

sarakhmen (changes requested):
Let's improve your solution ;)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
m-pnsk: It's a pity. I turned on my imagination for this(

sarakhmen (approved):
Good job ;)

PR #620, opened 2023-05-25
Ready for check

AnatolyNz:


okuzan (changes requested):
Set up a complete valid user in `setup` method, which will be called before each test, and then in the negative tests break the property you need - rewrite its password with a too-short one, etc.

src/main/java/core/basesyntax/service/RegistrationService.java
```diff
@@ -3,5 +3,6 @@
 import core.basesyntax.model.User;
 
 public interface RegistrationService {
-    User register(User user);
+    User register(User user) throws RegistrationException,
+            core.basesyntax.service.RegistrationException;
```
Ivan95kos, line 7:
```suggestion
    User register(User user) throws RegistrationException;
```

Ivan95kos (approved):
Great job, with minor comments

PR #621, opened 2023-05-27
Hw unit test

RostyslavOnysh:


olekskov (approved):
Good job overall, approved.
The only remark is that userId shouldn't be passed by user on registration or authentication stage so related validation and it's tests are redundant. Please consider for further tasks

aaananas (approved):

PR #642, opened 2023-06-27
Reg serv solution

knyazkyylubomir:


Ivan95kos (changes requested):
Please, fix all checks
Знімок екрана 2023-06-27 о 21 00 07

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,74 @@
...
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_SIZE_OF_STRING = 6;
+    private static final int MIN_ALLOWED_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        registerNullUser(user);
```
kozhukhovsky, line 15:
```suggestion
        checkNullUser(user);
```

Rename other methods as well, because we are checking them, not registering.

src/main/java/core/basesyntax/dao/StorageDaoImpl.java
```diff
@@ -2,20 +2,24 @@
 
 import core.basesyntax.db.Storage;
 import core.basesyntax.model.User;
+import java.util.ArrayList;
+import java.util.List;
 
 public class StorageDaoImpl implements StorageDao {
```
kozhukhovsky, line 8: There is no need to change anything, return as it was :)

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -5,5 +5,17 @@
 import java.util.List;
 
 public class Storage {
```
kozhukhovsky, line 7: There is no need to change anything, return as it was :)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,74 @@
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exception.InvalidDataException;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_SIZE_OF_STRING = 6;
```
kozhukhovsky, line 9: Not a very clear name. It's better to make a second variable, and write like the minimum length of the password, the minimum length of the login.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,116 @@
...
+    private Storage storage = new Storage();
+    private User bob;
+
+    @BeforeEach
+    void setUp() {
+        bob = new User("bob423", "123456", 19);
+    }
+
+    @AfterEach
+    void tearDown() {
```
kozhukhovsky, line 28: If you change the code back, you will need to call storage clear

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,116 @@
...
+        new Storage();
+    }
+
+    @Test
+    void register_registerUsersByCorrectData_ok() {
```
kozhukhovsky, line 33: You are not calling the register method, and we need to check that this logic works correctly.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,116 @@
...
+    @Test
+    void register_addUserByInvalidLogin_notOk() {
+        bob.setLogin("bob");
+        assertThrows(InvalidDataException.class, () -> {
+            registrationService.register(bob);
+        });
+    }
+
+    @Test
+    void register_addUserByInvalidPassword_notOk() {
```
kozhukhovsky, line 61: It is not clear why this particular password is invalid. It is better to check for edge cases, for example, that a value of 5 does not pass, and in the next test that 6 or more already passes.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,116 @@
...
+        storageDao.add(bob);
+        bob.setLogin("bob_jenkins");
+        storageDao.add(bob);
+        int actual = getPeople().size();
+        int expected = 3;
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    void register_getUserByExistLogin_ok() {
```
kozhukhovsky, line 45: No need to test how our dao methods work. The task of the tests here is to cover all possible variants of correct and incorrect registration.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,116 @@
...
+        storageDao.add(bob);
+        String actual = storageDao.get(bob.getLogin()).getLogin();
+        String expected = bob.getLogin();
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    void register_addUserByInvalidLogin_notOk() {
```
kozhukhovsky, line 53: From the name it is not clear why the user is not valid. Write what type check the minimum length of the login. And the same rules as described for the password.

kozhukhovsky (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
knyazkyylubomir: Thanks! clear() is working)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,75 @@
...
+        checkNullPassword(user);
+        checkNullAge(user);
+        checkAlreadyCreatedLogin(user);
+        checkLoginLessThan6Chars(user);
+        checkPasswordLessThan6Chars(user);
+        checkNotAllowedAge(user);
+        return storageDao.add(user);
+    }
+
+    public void checkNullUser(User user) {
```
Ivan95kos, line 27: For the future, methods that use only in class should be private 
```suggestion
    private void checkNullUser(User user) {
```

knyazkyylubomir: ok I forgot about that. 'facepalm' ))

PR #643, opened 2023-06-29
Completed initial task

AntonU91:


src/main/java/core/basesyntax/utility/RegistrationException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.utility;
+
+public class RegistrationException extends RuntimeException {
```
okuzan, line 3: The class `RegistrationException` is currently extending `RuntimeException` which is an unchecked exception. If your application should catch exceptions, consider making `RegistrationException` extend from `Exception` (a checked exception) instead. However, if these exceptions are something that will cause your application to stop, then it is alright to use `RuntimeException`.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,63 @@
...
+            throw new RegistrationException("User password should not be null");
+        }
+    }
+
+    private void validateUserAge(User user) {
+        if (user.getAge() < MIN_AGE) {
+            throw new RegistrationException("User age should not be less than "
+                    + MIN_AGE + " ages");
+        }
+        if (user.getAge() == null) {
```
okuzan, line 45: The null check for the user age should be performed before checking its value. Move this check to be the first in the 'validateUserAge' method.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,63 @@
...
+            throw new RegistrationException("User age should has null/empty value ");
+        }
+        if (user.getAge() < 0) {
+            throw new RegistrationException("User age should not be negative number ");
+        }
+    }
+
+    private void validateUserLogin(User user) {
+        if (user.getLogin() == null) {
```
okuzan, line 54: The null check for the user login should be performed before checking its length or if it's empty. Move this check to be the first in the 'validateUserLogin' method.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,63 @@
...
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new RegistrationException("User should not be null");
+        }
+        validateUserLogin(user);
+        validateUserAge(user);
+        validateUserPassword(user);
+        if (storageDao.get(user.getLogin()) == null) {
+            storageDao.add(user);
```
okuzan, line 22: If a user with the same login already exists in the storage, the method should throw an exception instead of adding a new user.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,63 @@
...
+
+    private void validateUserPassword(User user) {
+        if (user.getPassword().length() < MIN_SYMBOLS) {
+            throw new RegistrationException("User password should not be less than "
+                    + MIN_SYMBOLS + " symbols");
+        }
+        if (user.getPassword().isEmpty()) {
+            throw new RegistrationException("User password should not be empty");
+        }
+        if (user.getPassword() == null) {
```
okuzan, line 35: think about the order of those checks. What will happen if you try to call  `.isEmpty` , `.length()` on null?

The null check for the user password should be performed before checking its length or if it's empty. Move this check to be the first in the 'validateUserPassword' method.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+        initialUser.setAge(-12);
+        assertThrows(RegistrationException.class, () -> registrationService.register(initialUser));
+    }
+
+    @Test
+    void register_ageIsEqualTo18YearsOrMore_ok() {
+        initialUser.setLogin("dummy@37");
+        initialUser.setAge(18);
+        initialUser.setPassword("password1234");
+        User userToCheck = registrationService.register(initialUser);
```
okuzan, line 94: You are not asserting against the actual result of the registrationService.register method. Consider adding an assertEquals to compare the expected and actual User object returned by the register method.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static RegistrationService registrationService;
+    private static List<User> storage;
```
okuzan, line 19: The 'storage' variable is declared but not used. You should remove it if you don't plan to use it in this test class.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+    private static User initialUser;
+    private static User properToRegistUser;
+
+    @BeforeAll
+    static void initializeRegistrationService() {
+        registrationService = new RegistrationServiceImpl();
+        storage = new ArrayList<>();
```
okuzan, line 26: The 'storage' variable is initialized but never used. Consider removing this line.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+        initialUser.setLogin("dummy");
+        assertThrows(RegistrationException.class, () -> registrationService.register(initialUser));
+    }
+
+    @Test
+    void register_loginHasSixOrMoreSymbols_ok() {
+        initialUser.setLogin("dummy@37");
+        initialUser.setAge(18);
+        initialUser.setPassword("password1234");
+        User userToCheck = registrationService.register(initialUser);
```
okuzan, line 66: You are not asserting against the actual result of the registrationService.register method. Consider adding an assertEquals to compare the expected and actual User object returned by the register method.
image

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,63 @@
...
+        if (user.getLogin() == null) {
+            throw new RegistrationException("User login should not be null value ");
+        }
+        if (user.getLogin().length() < MIN_SYMBOLS) {
+            throw new RegistrationException("User login should not be less than "
+                    + MIN_SYMBOLS + " symbols");
+        }
+        if (user.getLogin().isEmpty()) {
+            throw new RegistrationException("User login should not be empty");
+        }
```
okuzan, line 63: your MIN_SYMBOLS is greater than 0, that means isEmpty check is basically unreachable

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static RegistrationService registrationService;
+    private static List<User> storage;
+    private static User initialUser;
+    private static User properToRegistUser;
```
okuzan, line 21: one valid user is enough

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Test
+    void register_loginLessThanSixSymbols_notOk() {
+        initialUser.setLogin("dummy");
+        assertThrows(RegistrationException.class, () -> registrationService.register(initialUser));
+    }
+
+    @Test
+    void register_loginHasSixOrMoreSymbols_ok() {
```
okuzan, line 62: one _ok test of a user which has all valid data should be enough. You basically have the same test twice, only named differently

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+        registrationService.register(properToRegistUser);
+        assertEquals(expected, Storage.people.size());
+    }
+
+    @Test
+    void register_userIsNotExistInStorage_addToStorage() {
+        int expected = Storage.people.size() + 1;
+        registrationService.register(properToRegistUser);
+        assertEquals(expected, Storage.people.size());
+    }
```
okuzan, line 136: combine with the positive test of registration, no need to have a separate test for that

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+        assertThrows(RegistrationException.class, () -> registrationService.register(initialUser));
+    }
+
+    @Test
+    void register_userIsAlreadyExistInStorage_notAddToStorage() {
+        Storage.people.add(properToRegistUser);
+        int expected = Storage.people.size();
+        registrationService.register(properToRegistUser);
+        assertEquals(expected, Storage.people.size());
+    }
```
okuzan, line 129: exception should be thrown

okuzan (changes requested):
edge cases not fully tested
image

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
AntonU91: Done

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
AntonU91: Done

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
AntonU91: Done

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
AntonU91: Done

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
AntonU91: Done

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
AntonU91: Done

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Ivan95kos: not fixed

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Ivan95kos: not fixed

Ivan95kos (changes requested):
Please, fix previous comments

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
AntonU91: Done

kozhukhovsky (approved):

PR #649, opened 2023-07-06
created UserTest;

kaptetos:
registerUser_validData_noExistingUser_success: This test creates two users, one of which already exists in the store (has the same username). The test checks that the new user successfully validates and is not equal to an existing user.

registerUser_shortLogin_exceptionThrown: This test creates a user with a login that is too short (less than 6 characters). The test verifies that the user throws a custom unchecked InvalidDataException during validation.

registerUser_shortPassword_exceptionThrown: This test creates a user with a password that is too short (less than 6 characters). The test verifies that the user throws a custom unchecked InvalidDataException during validation.

registerUser_underageUser_exceptionThrown: This test creates an underage user (under age 18). The test verifies that the user throws a custom unchecked InvalidDataException during validation.
registerUser_nullLogin_exceptionThrown: Checks that attempting to register a user with a null login throws an InvalidDataException.
registerUser_nullPassword_exceptionThrown: Checks that attempting to register a user with a null password throws an InvalidDataException.
registerUser_negativeAge_exceptionThrown: Checks that attempting to register a user with a negative age will throw an InvalidDataException.
registerUser_upperAgeLimit_validData_success: Checks that attempting to register a user with the maximum allowed age value (eg 150) does not throw an exception and the validation succeeds.

### DIFF
--- a/src/test/java/core/basesyntax/model/UserTest.java
+++ b/src/test/java/core/basesyntax/model/UserTest.java
@@ -1,0 +1,104 @@
+package core.basesyntax.model;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class UserTest {
+
+    @Test
+    void registerUser_validData_noExistingUser_success() {
+        User existingUser = createUser("john_doe", "password", 25);
+        User newUser = createUser("jane_doe", "newpassword", 20);
+
+        Assertions.assertNotEquals(existingUser, newUser);
+    }
+
+    @Test
+    void registerUser_shortLogin_exceptionThrown() {
+        User user = createUser("abc", "password", 25);
+
+        Assertions.assertThrows(InvalidDataException.class, () -> {
+            validateUser(user);
+        });
+    }
+
+    @Test
+    void registerUser_shortPassword_exceptionThrown() {
+        User user = createUser("john_doe", "abc", 25);
+
+        Assertions.assertThrows(InvalidDataException.class, () -> {
+            validateUser(user);
+        });
+    }
+
+    @Test
+    void registerUser_underageUser_exceptionThrown() {
+        User user = createUser("john_doe", "password", 17);
+
+        Assertions.assertThrows(InvalidDataException.class, () -> {
+            validateUser(user);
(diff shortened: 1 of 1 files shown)


kozhukhovsky (changes requested):
You must follow the assignment. We don't need to test user creation. And create a user registration service with the necessary checks. And then tests cover all possible options.

src/test/java/core/basesyntax/model/UserTest.java
```diff
@@ -0,0 +1,104 @@
+package core.basesyntax.model;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class UserTest {
```
olekskov, line 6: Please remove this class. It tests itself and has no connection to code of program

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -9,6 +10,22 @@ public class RegistrationServiceImpl implements RegistrationService {
...
+        }
+        if (user.getPassword() == null || user.getPassword().length() < 6) {
+            throw new InvalidUserException("Invalid password: " + user.getPassword());
+        }
+        if (user.getAge() == null || user.getAge() < 18) {
+            throw new InvalidUserException("Invalid age: " + user.getAge());
+        }
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidUserException("User with login already exists" + user.getLogin());
+        }
```
olekskov, line 29: 1. Please extract magic values to constants with proper naming
2. Please make exception messages hold information to determine what is wrong with login or pass. Message "Invalid login: john" doesn't give any clue to user that he should use at least 6 symbols

olekskov (changes requested):

src/test/java/core/basesyntax/model/UserTest.java
kaptetos: i removed this class , place it i created RegistrationServiceTest

Ivan95kos (approved):
Great job

PR #665, opened 2023-07-21
Implement register(User user) method in RegistrationServiceImpl class

Dimagaa:
Write test for RegistrationServiceImpl class

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -5,10 +5,49 @@ import core.basesyntax.dao.StorageDaoImpl;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int PASSWORD_MIN_LENGTH = 6;
+    private static final int LOGIN_MIN_LENGTH = 6;
+    private static final int MIN_AGE = 18;
+
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new InvalidUserDataException("Provided User is null");
+        }
+        validateLogin(user.getLogin());
+        validatePassword(user.getPassword());
+        validateAge(user.getAge());
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidUserDataException(
+                    "User with such login has been already registered. Login: "
+                    + user.getLogin());
+        }
+        return storageDao.add(user);
     }
+
+    private void validateLogin(String login) {
+        if (login == null || login.length() < LOGIN_MIN_LENGTH) {
+            throw new InvalidUserDataException("Invalid User login was provided: "
+                    + login + ". Login can't be null and be less than "
+                    + LOGIN_MIN_LENGTH + " characters");
+        }
+    }
+
+    private void validatePassword(String password) {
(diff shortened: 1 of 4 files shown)


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,49 @@
...
+        }
+        validateLogin(user.getLogin());
+        validatePassword(user.getPassword());
+        validateAge(user.getAge());
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidUserDataException(
+                    "User with such login has been already registered. Login: "
+                    + user.getLogin());
+        }
+        return storageDao.add(user);
```
Den-moroz, line 27: I would first add the user to the database and then return it from it, so you will know for sure that your user was added.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,49 @@
...
+                    + password + ". Password can't be null and be less than "
+                    + PASSWORD_MIN_LENGTH + " characters");
+        }
+    }
+
+    private void validateAge(Integer age) {
+        if (age == null || age < MIN_AGE) {
+            throw new InvalidUserDataException("Invalid User age was provided: "
+                    + age + ". Age can't be null or less then " + MIN_AGE);
+        }
```
Den-moroz, line 50: Don’t create one big if with all conditions
image

Den-moroz (changes requested):
Good job! Let's improve, also you could do this task without using stream API.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
Dimagaa: When we use add(User user) method from StorageDao class, we have to get added User with id, that User obtained when was added, we have logic with a bug otherwise (this case covered by tests)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
Dimagaa: In the 'bad example' we have overloaded 'if', which contains checks for all fields. In my solution, I have divided logic and a specific informative exception message for each field. Also, consider, that 'good example' doesn't cover all possible cases. Therefore we shouldn't just copy the example.

Den-moroz: I meant that you could display two messages about null and about the length, but you can leave it as it was.

Den-moroz (approved):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
Den-moroz: There will be no error, I just checked that.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
Dimagaa: Got it and changed the conditions. Thanks!

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
Dimagaa: Look at my logs and what will happen if there is no such user in Storage after the register method execution.

Screenshot 2023-07-22 203523

Vasul-Pankiv (approved):
So good solution 🚀

PR #668, opened 2023-07-21
implemented and tested 'register' method

DmytroV95:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,51 @@
...
+        if (user.getAge() == null) {
+            throw new UserRegistrationException("User age "
+                    + user.getAge() + " is not specified. Age can't be null");
+        }
+        if (user.getAge() < USER_MIN_ACCESSIBLE_AGE) {
+            throw new UserRegistrationException("User age "
+                    + user.getAge()
+                    + " less than " + USER_MIN_ACCESSIBLE_AGE + " years old");
+        }
+        return storageDao.add(user);
```
Denis-Balako, line 49: I recommend you separate the registration logic to check methods and adding to storage

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final String USER_PASSWORD_LENGTH_6_SYMBOLS_OK = "bob123";
+
+    @BeforeAll
+    static void beforeAll() {
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @ParameterizedTest
+    @MethodSource("register_userData_Ok")
+    void register_addUserToDatabase_Ok(User input) {
```
Denis-Balako, line 32:
```suggestion
    void register_addUserToDatabase_ok(User input) {
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+        return Stream.of(
+                Arguments.of(new User(
+                        USER_LOGIN_LENGTH_6_SYMBOLS_OK,
+                        USER_PASSWORD_LENGTH_6_SYMBOLS_OK,
+                        USER_MIN_ACCESSIBLE_AGE)),
+                Arguments.of(new User(
+                        USER_LOGIN_LENGTH_6_SYMBOLS_OK,
+                        USER_PASSWORD_LENGTH_6_SYMBOLS_OK,
+                        USER_AGE_MORE_THAN_18_OK))
+        );
```
Denis-Balako, line 55: I'm far from streams, but how are adding 2 users with the same login? We expect an exception for this case.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+class RegistrationServiceImplTest {
+    private static RegistrationServiceImpl registrationService;
+    private static final int USER_AGE_LESS_18_NOT_OK = 17;
+    private static final int USER_AGE_MORE_THAN_18_OK = 30;
+    private static final String USER_LOGIN_LENGTH_LESS_6_SYMBOLS_NOT_OK = "Bob";
+    private static final String USER_LOGIN_LENGTH_6_SYMBOLS_OK = "Bob123";
+    private static final String USER_PASSWORD_LENGTH_LESS_6_SYMBOLS_NOT_OK = "bob";
+    private static final String USER_PASSWORD_LENGTH_6_SYMBOLS_OK = "bob123";
```
Denis-Balako, line 23: Try to test for user = null, 2 valid different users, and 2 valid similar users.

Denis-Balako (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
DmytroV95: This construct provides us with a stream of new users (eg a list of different users). Each part of the stream that starts with "Arguments.of" is a separate user. And we add them one by one to test the method and test adding a new user with the specified data to the repository. We don't add them all at once. It's like we can create every test method for every test case. But this design allows you to accumulate all test cases it in one place.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
DmytroV95: Adding 2 valid different users is not necessary because if we cover all possible cases during other test. To my mind it will be valid if we add 2 users too.

src/main/java/core/basesyntax/model/User.java
```diff
@@ -8,6 +8,12 @@ public class User {
     private String password;
     private Integer age;
 
+    public User(String login, String password, Integer age) {
+        this.login = login;
+        this.password = password;
+        this.age = age;
+    }
+
```
Skarbey, line 16: I have doubts that this is an unnecessary implementation.

src/main/java/core/basesyntax/model/User.java
```diff
@@ -58,4 +64,12 @@ public boolean equals(Object o) {
...
         return Objects.hash(login, password, age);
     }
+
+    @Override
+    public String toString() {
+        return "User{"
+                + ", login='" + login + '\''
+                + ", password='" + password + '\''
+                + ", age=" + age + '}';
+    }
```
Skarbey, line 74: and this may too

src/main/java/core/basesyntax/model/User.java
DmytroV95: Thank you for comment. Constructor with arguments used in my data test sets. In my opinion, its use is reasonable.

src/main/java/core/basesyntax/model/User.java
DmytroV95: Thank you for comment. Yes, maybe a little bit. That is addition method override i used to display of user data during debug process. i think its reasonable to use in our study process)) If you don't mind i will ask mentors in which cases we need use it.

Skarbey (approved):

Skarbey:
I like how did you separate checks by methods in your code!

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,127 @@
...
+                        null)),
+                Arguments.of(new User(null, null, null))
+        );
+    }
+}
+
+
+
+
+
```
Denis-Balako, line 127: don't forget to delete redundant lines at the end of the file.

Denis-Balako (approved):
Great job!

pom.xml
```diff
@@ -77,4 +83,4 @@
             </plugins>
         </pluginManagement>
     </build>
-</project>
+</project>
```
Vasul-Pankiv, line 86: add empty line to the end of file

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,127 @@
...
+    void register_addUserToDatabase_ok(User input) {
+        User actualResult = registrationService.register(input);
+        assertEquals(input.getLogin(), actualResult.getLogin());
+        assertEquals(input.getPassword(), actualResult.getPassword());
+        assertEquals(input.getAge(), actualResult.getAge());
+    }
+
+    @ParameterizedTest
+    @MethodSource("register_userData_NotOk")
+    void register_addUserToDatabase_notOk(User input) {
```
Vasul-Pankiv, line 45: why notOkay? (specify method name)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,127 @@
...
+        assertThrows(UserRegistrationException.class, () -> registrationService.register(input));
+    }
+
+    @Test
+    void register_addNullUserToDatabase_notOk() {
+        assertThrows(UserRegistrationException.class, () -> registrationService.register(null));
+    }
+
+    @Test
+    void register_addEqualsUsersToDatabase_notOk() {
```
Vasul-Pankiv, line 55:
```suggestion
    void register_addExistingUser_notOk() {
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,127 @@
...
+        User user = new User(
+                USER_LOGIN_LENGTH_6_SYMBOLS_OK,
+                USER_PASSWORD_LENGTH_6_SYMBOLS_OK,
+                USER_MIN_ACCESSIBLE_AGE);
+        storageDao.add(user);
+        assertThrows(UserRegistrationException.class, () -> registrationService.register(user));
+    }
+
+    private static Stream<Arguments> register_userData_ok() {
```
Vasul-Pankiv, line 64:
```suggestion
    private static Stream<Arguments> register_validUser_ok() {
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,127 @@
...
+                        USER_PASSWORD_LENGTH_6_SYMBOLS_OK,
+                        USER_MIN_ACCESSIBLE_AGE)),
+                Arguments.of(new User(
+                        USER_LOGIN_LENGTH_6_SYMBOLS_OK,
+                        USER_PASSWORD_LENGTH_6_SYMBOLS_OK,
+                        USER_AGE_MORE_THAN_18_OK))
+        );
+    }
+
+    private static Stream<Arguments> register_userData_NotOk() {
```
Vasul-Pankiv, line 77:
```suggestion
    private static Stream<Arguments> register_invalidUser_NotOk() {
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Vasul-Pankiv: not resolved)

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,66 @@
...
+
+    private void validateUserLogin(User user) {
+        if (user.getLogin() == null || user.getLogin().equals(WHITE_SPACE)) {
+            throw new UserRegistrationException("User login "
+                    + user.getLogin() + " is not specified. Login can't be null");
+        }
+        if (user.equals(storageDao.get(user.getLogin()))) {
+            throw new UserRegistrationException("User by specified login "
+                    + user.getLogin() + " already exist");
+        }
```
nacenik, line 59: better to do this check after validation `serLogin.length()` cause you check all input data that is not null, but why should you do it if the login cannot be less than 6 characters?

nacenik (changes requested):
Good solution, let's improve it :)

nacenik (approved):
LGTM! 👍

PR #672, opened 2023-07-23
Implemented register method and unit tests

OlegTuryk:


src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,83 @@
...
+    void setUp() {
+        registrationService = new RegistrationServiceImpl();
+        storageDao = new StorageDaoImpl();
+        storageDao.add(EXISTING);
+    }
+
+    @Test
+    void validRegistration() {
+        User user = new User(VALID_LOGIN, VALID_PASSWORD, VALID_AGE);
+        User registeredUser = registrationService.register(user);
```
Someboty, line 36: Your register method returns original user. You don't check if user was added to db

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Test
+    void nullLoginRegistration() {
+        User user = new User(null, VALID_PASSWORD, VALID_AGE);
+        assertThrows(RegistrationServiceException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void invalidLoginLengthRegistration() {
```
Someboty, line 61: Why it's invalid? Specify it in test's name

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Test
+    void nullPasswordRegistration() {
+        User user = new User(VALID_LOGIN, null, VALID_AGE);
+        assertThrows(RegistrationServiceException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void invalidPasswordLengthRegistration() {
```
Someboty, line 73: Same problem

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,83 @@
...
+        User user = new User(VALID_LOGIN, INVALID_PASSWORD, VALID_AGE);
+        assertThrows(RegistrationServiceException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void invalidAgeRegistration() {
```
Someboty, line 79: And here

Someboty (changes requested):
Nice work. But let's add some improvements to it.

Also, I think you should add more tests:
"In your tests very important to check all possible edge cases. For example, we have min password length 6 characters. In this case would be good to test: empty password (0-length - not-ok), ‘abc’ (3-length - not-ok), ‘abcdf’ (5-length - not-ok edge case), ‘abcdef’ (6-length - ok edge case), ‘abcdefgh’ (8-length - ok). Few not-ok checks you can have in one not-ok test method. Same for ok-test. Testing only ‘abc’ (3-length - not-ok) and ‘abcdefgh’ (8-length - ok) is not enough because we do not check how the program behave with edge cases"

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,83 @@
...
+        User user = new User(VALID_LOGIN, VALID_PASSWORD, VALID_AGE);
+        User registeredUser = registrationService.register(user);
+        assertNotNull(registeredUser.getId());
+        assertEquals(user.getLogin(), registeredUser.getLogin());
+        assertEquals(user.getPassword(), registeredUser.getPassword());
+        assertEquals(user.getAge(), registeredUser.getAge());
+    }
+
+    @Test
+    void nullUserRegistration() {
```
holsoni, line 44: When you are expecting an exception to be thrown, you can add _notOk at the end of methdos. It means that you are trying to do some not appropriate actions and you expect an exception to be thrown.

holsoni (changes requested):
Good solution, one comment!

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+import core.basesyntax.RegistrationServiceException;
+import core.basesyntax.dao.StorageDao;
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceTest {
+    private static final String INVALID_LOGIN = "LOGIN";
```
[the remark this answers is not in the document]
OlegTuryk: Invalid - не валідний

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @BeforeEach
+    void setUp() {
+        registrationService = new RegistrationServiceImpl();
+        storageDao = new StorageDaoImpl();
+        storageDao.add(EXISTING);
+    }
+
+    @Test
+    void validRegistration() {
```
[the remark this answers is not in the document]
OlegTuryk: I don't think so

Someboty (approved):

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceTest {
+    private static final String INVALID_LOGIN = "LOGIN";
+    private static final String VALID_LOGIN = "LOGIN@";
+    private static final String EXISTING_LOGIN = "LOGIN_OLEG";
+    private static final String INVALID_PASSWORD = "PASS";
+    private static final String VALID_PASSWORD = "PASSWO";
+    private static final int INVALID_AGE = 15;
```
nacenik, line 20: The same about age. What happens if I change minimum age from 18 to 17 or 16?

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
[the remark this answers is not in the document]
nacenik: Your test name doesn't tell much information. Valid registration - what is the method name that you test? What is the expected result? Fix for all test names 
image

nacenik (changes requested):
Let's improve your solution

Vasul-Pankiv (approved):

PR #676, opened 2023-07-24
Implemented task

OleksandrSaven:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,122 @@
...
+        });
+    }
+
+    @Test
+    void register_minLogin_notOk() {
+        user.setLogin("Osa");
+        assertThrows(RegistrationException.class, () -> {
+            registrationService.register(user);
+        });
+    }
```
Zel1oy, line 76: It would be better to insert some messages to test cases, so that user would know what's wrong

Zel1oy (changes requested):

nacenik (changes requested):
Good solution, let's improve it

nacenik (approved):
LGTM! 👍

PR #701, opened 2023-08-05
The RegistrationServiceImpl class has been implemented. Added checkUs…

Olga-Melnichenko:
…erLogin, checkUserPassword, checkUserAge. Implemented class RegistrationServiceImplTest

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -5,10 +5,42 @@ import core.basesyntax.dao.StorageDaoImpl;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_AGE = 18;
+    private static final int MIN_PASSWORD_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        checkUserLogin(user.getLogin());
+        checkUserPassword(user.getPassword());
+        checkUserAge(user.getAge());
+        return storageDao.add(user);
+    }
+
+    private void checkUserLogin(String login) {
+        if (login == null || login.isEmpty() || login.isBlank()) {
+            throw new RuntimeException("User login is null");
+        }
+        if (storageDao.get(login) != null) {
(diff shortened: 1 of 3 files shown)


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,42 @@
...
+            throw new RuntimeException("User login is null");
+        }
+        if (storageDao.get(login) != null) {
+            throw new RuntimeException("User with this login exists in storage");
+        }
+    }
+
+    private void checkUserPassword(String password) {
+        if (password == null || password.isEmpty() || password.isBlank()) {
+            throw new RuntimeException("User password is null");
```
olekskov, line 31: Statement that it's null is not fully correct, better to have more generic message in this case
```suggestion
            throw new RuntimeException("User password wasn't provided");
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,42 @@
...
-        return null;
+        checkUserLogin(user.getLogin());
+        checkUserPassword(user.getPassword());
+        checkUserAge(user.getAge());
+        return storageDao.add(user);
+    }
+
+    private void checkUserLogin(String login) {
+        if (login == null || login.isEmpty() || login.isBlank()) {
+            throw new RuntimeException("User login is null");
```
olekskov, line 22: same here

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,118 @@
...
+    @Test
+    void register_newUser_Ok() {
+        User newUser = registrationService.register(user);
+        assertEquals(user, newUser);
+    }
+
+    @Test
+    void register_newUserSameInformation_notOk() {
+        assertEquals(user, registrationService.register(user));
+        assertThrows(RuntimeException.class, () -> registrationService.register(user));
```
olekskov, line 40: In general it's better to also check that exception messages complies with expected to avoid cases where other validation accidentally worked instead of the one that is tested

olekskov (approved):
Nice job, approved.
Small comments below are for further reference

PR #704, opened 2023-08-08
Version1

stas83ggg:


overpathz (approved):
Approved.

But pay attention to some minor things:
1) There is no such requirement in homework, but what if we got a null user? Should we do this check?
2) I would split one variable MIN_LENGTH into -> MIN_LOGIN_LENGTH and MIN_PASSWORD_LENGTH since in some cases we can change the requirement of length separately for both
3) if you have repeated values in tests, move it to constants

PR #706, opened 2023-08-14
Has written solution for task jv-registration-validation

Stelmaschyk:


src/main/java/core/basesyntax/model/RegistrationServiceImpl.java
```diff
@@ -0,0 +1,40 @@
+package core.basesyntax.model;
+
+import core.basesyntax.dao.StorageDao;
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exception.RegistrationException;
+import core.basesyntax.service.RegistrationService;
+
+public class RegistrationServiceImpl implements RegistrationService {
+    static final int AGE_PERMISION = 18;
+    static final int MIN_PASSWORD_LENGTH = 5;
```
okuzan, line 10: a bit confusing to have not-ok-edge case as a constant, I'd make that 6 and changed condition from <= to <

src/main/java/core/basesyntax/model/RegistrationServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+    private final StorageDao storageDao = new StorageDaoImpl();
+
+    @Override
+    public User register(User user) {
+        if (user.getPassword() == null) {
+            throw new RegistrationException("Password can't be null");
+        }
+        if (user.getPassword().length() <= MIN_PASSWORD_LENGTH) {
+            throw new RegistrationException("Password length less then 6(six) characters");
```
okuzan, line 19: don't hardcode limits, rely on your constants

src/main/java/core/basesyntax/model/RegistrationServiceImpl.java
okuzan, line 10: add constant for login length too (better option), or change the naming and have one for both

src/main/java/core/basesyntax/model/RegistrationServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+        }
+        if (user.getLogin().length() <= MIN_PASSWORD_LENGTH) {
```
okuzan, line 21: check if get login isn't null first

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,160 @@
...
+    @Test
+    void storageDao_addUserToStorage_ok() {
+        User user = new User();
+        user.setId(1L);
+        user.setPassword("1234562");
+        user.setLogin("Cat");
+        user.setAge(19);
+        storageDao.add(user);
+        User actual = storageDao.get("Cat");
+        assertEquals(user, actual);
```
okuzan, line 64: Tests in this class should test methods of registration service, you basically test storageDao

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Test
+    void register_loginLengthMoreFiveCharacters_Ok() {
+        User user = new User();
+        user.setLogin("123456");
+        user.setPassword("userpassword");
+        int actual = user.getLogin().length();
+        int expectLoginLength = 6;
+        assertTrue(expectLoginLength <= actual);
+    }
```
okuzan, line 53: only one _ok test is necessary, when every field is valid

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,160 @@
...
+        user.setPassword("password");
+        user.setLogin("Petrenko");
+        user.setAge(17);
+        assertThrows(RegistrationException.class, () -> {
+            registrationService.register(user);
+        });
+    }
+
+    @Test
+    void register_AgeMoreThan17_Ok() {
```
okuzan, line 150: in some tests, you have uppercase after _ underscores, in some you have lowercase. Please unify your naming

src/main/java/core/basesyntax/model/RegistrationServiceImpl.java
```diff
@@ -0,0 +1,46 @@
+package core.basesyntax.model;
+
+import core.basesyntax.dao.StorageDao;
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exception.RegistrationException;
+import core.basesyntax.service.RegistrationService;
+
+public class RegistrationServiceImpl implements RegistrationService {
+    static final int AGE_PERMISION = 18;
```
overpathz, line 9:
```suggestion
    private static final int MIN_AGE = 18;
```

overpathz: Not neccessary to change the name of the variable(but consider to, since age permission sound like smth different from min age), but you should change to private

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,170 @@
...
+    private static StorageDao storageDao;
+
+    @BeforeEach
+    void setUp() {
+        registrationService = new RegistrationServiceImpl();
+        storageDao = new StorageDaoImpl();
+    }
+
+    @Test
+    void register_nullPasswordValue_notOk() {
```
overpathz, line 27: Consider to set null explicitly to make test more clear

src/main/java/core/basesyntax/model/RegistrationServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+        }
+        if (user.getLogin().length() < MIN_PASSWORD_AND_LOGIN_LENGTH) {
+            throw new RegistrationException("Login length less then "
+                    + MIN_PASSWORD_AND_LOGIN_LENGTH + " characters");
+        }
+        if (user.getLogin().equals(storageDao.get(user.getLogin()))) {
+            throw new RegistrationException("This user exist in storage, please change your "
+                    + "login");
+        }
+        if (user.getAge() < AGE_PERMISION || user.getAge() == null) {
```
overpathz, line 33: if age would be a wrapper class, for instance, Integer or Long, comparing it to some number can cause NullPointerException. So you have to perform any check or change approach to checking the min age.

overpathz (changes requested):
Good, but need some fixes

overpathz (approved):

PR #716, opened 2023-08-28
implemented unit testing for user registration

RedArtelerist:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
-        return null;
+        if (user == null) {
+            throw new RegistrationException("User can't be null");
+        }
+        checkValidLogin(user.getLogin());
+        checkValidPassword(user.getPassword());
+        checkValidAge(user.getAge());
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new RegistrationException("User with this login already exists");
+        }
```
liliia-ponomarenko, line 24: Let’s  move it to checkValidLogin method

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
+        return storageDao.add(user);
+    }
+
+    private void checkValidLogin(String login) {
+        if (login == null) {
+            throw new RegistrationException("Login can't be null");
+        }
+        if (login.trim().length() < MIN_LOGIN_LENGTH) {
+            throw new RegistrationException("Login must be at least 6 characters");
```
liliia-ponomarenko, line 33: Use constant here. If you change the constant you also would need to change the message
```suggestion
            throw new RegistrationException("Login must be at least " +  MIN_LOGIN_LENGTH + " characters");
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
+            throw new RegistrationException("User with this login already exists");
+        }
+        return storageDao.add(user);
+    }
+
+    private void checkValidLogin(String login) {
+        if (login == null) {
+            throw new RegistrationException("Login can't be null");
+        }
+        if (login.trim().length() < MIN_LOGIN_LENGTH) {
```
liliia-ponomarenko, line 32: If you want use trim here you also shod parse input user credential before saving

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
+            throw new RegistrationException("Login must be at least 6 characters");
+        }
+    }
+
+    private void checkValidPassword(String password) {
+        if (password == null) {
+            throw new RegistrationException("Password can't be null");
+        }
+        if (password.trim().length() < MIN_PASSWORD_LENGTH) {
+            throw new RegistrationException("Password must be at least 6 characters");
```
liliia-ponomarenko, line 42: the same here

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Test
+    void register_nullLogin_notOk() {
+        user.setLogin(null);
+        assertThrows(RegistrationException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void register_loginLength_notOk() {
```
liliia-ponomarenko, line 52:
```suggestion
    void register_invalidLoginLength_notOk() {
```
check it everywhere

liliia-ponomarenko (changes requested):
good job! Let’s fix a few mistakes ;)

liliia-ponomarenko (changes requested):
triggered reject

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
RedArtelerist: If I understand it's better trim in setters?

Andriy-dz (approved):
Very good solution)

PR #719, opened 2023-09-01
'Registration-validation' task is implemented

Alexsandr-Yablunovskyi:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,211 @@
...
+        User actual = registrationService.register(user);
+        assertEquals(user, actual);
+    }
+
+    @Test
+    void register_ifLoginIsPresent_NotOk() {
+        User user = new User();
+        user.setLogin(VALID_LOGIN);
+        user.setPassword("AnyPassword");
+        user.setAge(20);
```
mmosii, line 53: Maybe will be better to make all this values constants, like common mistakes are telling us)
and in the rest of tests too.

mmosii (changes requested):
Good job, tests on all possible edge cases) 
minor issue with constants

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Alexsandr-Yablunovskyi: Good point! I thought about that too.
Fixed.

src/main/java/core/basesyntax/db/Storage.java
```diff
@@ -5,5 +5,5 @@
 import java.util.List;
 
 public class Storage {
-    public static final List<User> people = new ArrayList<>();
+    public static final List<User> PEOPLE = new ArrayList<>();
```
NazarSmith228, line 8: you should not modify existing classes - there is no need for that

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,66 @@
...
-        return null;
+        checkLogin(user);
+        checkPassword(user);
+        checkAge(user);
+        addOriginUser(user);
+        return user;
+    }
+
+    private static void checkLogin(User user) {
+        String exceptionMessage = "User login length should be at least 6 characters.";
```
NazarSmith228, line 21: there is no need to declare a separate variable for the exception message - you can inline it

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,66 @@
...
+        }
+        if (user.getLogin().isEmpty()) {
+            throw new RegistrationException(exceptionMessage);
+        }
+        if (user.getLogin().isBlank()) {
+            throw new RegistrationException(exceptionMessage);
+        }
+        if (user.getLogin().length() < 6) {
+            throw new RegistrationException(exceptionMessage);
+        }
```
NazarSmith228, line 33: there is no sense of using both `isEmpty` and `isBlank` if you need to just assert login length

I would suggest to unite `isBlank` and `lenght` conditions and use a message like `User login should have at least 6 non-whitespace characters`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,66 @@
...
     public User register(User user) {
-        return null;
+        checkLogin(user);
+        checkPassword(user);
+        checkAge(user);
+        addOriginUser(user);
+        return user;
+    }
+
+    private static void checkLogin(User user) {
```
NazarSmith228, line 20: why is it static?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,66 @@
...
+        if (user.getLogin().isBlank()) {
+            throw new RegistrationException(exceptionMessage);
+        }
+        if (user.getLogin().length() < 6) {
+            throw new RegistrationException(exceptionMessage);
+        }
+
+    }
+
+    private static void checkPassword(User user) {
```
NazarSmith228, line 37: why is it static?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,66 @@
...
+        }
+        if (user.getPassword().isBlank()) {
+            throw new RegistrationException(exceptionMessage);
+        }
+        if (user.getPassword().length() < 6) {
+            throw new RegistrationException(exceptionMessage);
+        }
+    }
+
+    private static void checkAge(User user) {
```
NazarSmith228, line 53: why is it static?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,66 @@
...
+        }
+        if (user.getPassword().isEmpty()) {
+            throw new RegistrationException(exceptionMessage);
+        }
+        if (user.getPassword().isBlank()) {
+            throw new RegistrationException(exceptionMessage);
+        }
+        if (user.getPassword().length() < 6) {
+            throw new RegistrationException(exceptionMessage);
+        }
```
NazarSmith228, line 50: the same remark as in `checkLogin` method

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,66 @@
...
+            throw new RegistrationException("User age should be 18 or more");
+        }
+    }
+
+    private void addOriginUser(User user) {
+        if (storageDao.get(user.getLogin()) == null) {
+            storageDao.add(user);
+        } else {
+            throw new RegistrationException("User with login \"" + user.getLogin()
+                    + "\" already exists in the storage");
```
NazarSmith228, line 64: you can use method `formatted` instead of concatenation

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,66 @@
...
+        if (user.getLogin() == null) {
+            throw new RegistrationException("Login can't be null.");
+        }
+        if (user.getLogin().isEmpty()) {
+            throw new RegistrationException(exceptionMessage);
+        }
+        if (user.getLogin().isBlank()) {
+            throw new RegistrationException(exceptionMessage);
+        }
+        if (user.getLogin().length() < 6) {
```
NazarSmith228, line 31: 6 seems to be a magic number - consider defining a constant for it

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,66 @@
...
+        if (user.getPassword().isBlank()) {
+            throw new RegistrationException(exceptionMessage);
+        }
+        if (user.getPassword().length() < 6) {
+            throw new RegistrationException(exceptionMessage);
+        }
+    }
+
+    private static void checkAge(User user) {
+        if (user.getAge() < 18) {
```
NazarSmith228, line 54: 18 seems to be a magic number - consider defining a constant for it

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+    private static final String MIN_EDGE_LENGTH_VALID = "7 chars";
+    private static final String MAX_EDGE_LENGTH_INVALID = "5char";
+    private static final String IS_EMPTY = "";
+    private static final String IS_BLANK = "      ";
+    private static final String VALID_PASSWORD = "ValidPassword";
+    private static final int VALID_AGE = 19;
+    private static final int MIN_VALID_AGE = 18;
+    private static final int MAX_INVALID_AGE = 17;
+    private static final int AGE_IS_ZERO = 0;
+    private static final int AGE_IS_NEGATIVE = -18;
```
NazarSmith228, line 28: I consider constants in tests a bad pattern

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+    private static RegistrationService registrationService;
+    private static User user;
+
+    @BeforeAll
+    static void beforeAll() {
+        registrationService = new RegistrationServiceImpl();
+        user = new User();
+    }
```
NazarSmith228, line 36: I don't think there is a necessity of using `@BeforeEach` with static variables. In addition - you don't need to declare some generic `User`, since you should prepare those users per each test separately - you're testing `RegistrationService` here

why not just:
```suggestion
    private final RegistrationService registrationService = new RegistrationServiceImpl();
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+    @BeforeEach
+    void setUp() {
+        user.setLogin(VALID_LOGIN);
+        user.setPassword(VALID_PASSWORD);
+        user.setAge(VALID_AGE);
+    }
```
NazarSmith228, line 43: due to my previous comment - you can remove this method

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,66 @@
...
+            throw new RegistrationException("User age should be 18 or more");
+        }
+    }
+
+    private void addOriginUser(User user) {
```
NazarSmith228, line 59: weird method name - what is `origin`?

how about `saveUser`?

also this method should not be `void` - `storageDao.add()` returns `User`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,66 @@
...
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        checkLogin(user);
+        checkPassword(user);
+        checkAge(user);
+        addOriginUser(user);
+        return user;
```
NazarSmith228, line 17:
```suggestion
        return saveUser(user);
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+    @AfterEach
+    void tearDown() {
+        Storage.PEOPLE.clear();
+
```
NazarSmith228, line 48: redundant blank line

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    @Test
+    void register_ifLoginIsAbsent_Ok() {
+        User actual = registrationService.register(user);
+        assertEquals(user, actual);
```
NazarSmith228, line 54: you should create a user, which you want to register directly in test

and then compare the returned user with the one, you created

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.PEOPLE.clear();
+
+    }
+
+    @Test
+    void register_ifLoginIsAbsent_Ok() {
```
NazarSmith228, line 52: I don't get this method name 

does it mean that if login is absent, the registration should be ok?

and this is the biggest problem of your test class - it's really hard to understand wha't going on

in fact you set a valid login in `@BeforeEach` - how should a reader like me know it? apart from that your test method name confuses the reader

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+    void register_LoginMaxEdgeInvalidLength_NotOk() {
+        user.setLogin(MAX_EDGE_LENGTH_INVALID);
+        assertThrows(RegistrationException.class, () -> registrationService.register(user),
+                "Login has invalid length. The RegistrationException should be thrown");
+    }
+
+    @Test
+    void register_PasswordValidLength_Ok() {
+        User actual = registrationService.register(user);
+        assertEquals(user.getPassword(), actual.getPassword());
```
NazarSmith228, line 121: in some tests you're comparing users, in others you're comparing passwords...

src/main/java/core/basesyntax/db/Storage.java
Alexsandr-Yablunovskyi: After "mvn clean package" I received:
Name 'people' must match pattern '^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$'. [ConstantName]
So I changed it. How can I solve this problem in another way?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,56 @@
...
+        checkPassword(user);
+        checkAge(user);
+        return saveUser(user);
+    }
+
+    private void checkLogin(User user) {
+        if (user.getLogin() == null) {
+            throw new RegistrationException("Login can't be null.");
+        }
+        if (user.getLogin().isBlank() || user.getLogin().length() < 6) {
```
liliia-ponomarenko, line 25:
```suggestion
        if (user.getLogin().isBlank() || user.getLogin().length() < MIN_VALID_LENGTH) {
```
Use constant here

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,56 @@
...
+            throw new RegistrationException(
+                    "User login should have at least 6 non-whitespace characters");
```
liliia-ponomarenko, line 27: Also better to use a constant for the message, if you change the constant value  you need to change the message too (check it in other messages)
```suggestion
                    "User login should have at least " + MIN_VALID_LENGTH + " non-whitespace characters");
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,56 @@
...
+        }
+
```
liliia-ponomarenko, line 29:
```suggestion
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,149 @@
...
+class RegistrationServiceImplTest {
+    private final RegistrationService registrationService = new RegistrationServiceImpl();
+    private final User user = new User();
+
+    @BeforeEach
+    void setUp() {
+        user.setLogin("ValidLogin");
+        user.setPassword("ValidPassword");
+        user.setAge(19);
+    }
```
liliia-ponomarenko, line 22: Let’s  remove this setUp and create separate testing  data for each  test

liliia-ponomarenko (changes requested):
Let’s  fix a few mistakes  ;)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @Test
+    void register_AgeIsNegative_NotOk() {
+        User user = new User();
+        user.setLogin("ValidLogin");
+        user.setPassword("ValidPassword");
+        user.setAge(-18);
+        assertThrows(RegistrationException.class, () -> registrationService.register(user));
+    }
```
Andriy-dz, line 173: I would also add tests for user = null

Andriy-dz (approved):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Alexsandr-Yablunovskyi: Thank you. Fixed

PR #720, opened 2023-09-01
implemented RegistrationServiceImpl class and test

mmosii:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final int VALID_AGE = 22;
+    private static final int YOUNGER_AGE = 17;
+    private static final int NEGATIVE_AGE = -20;
+    private static final String NULL_VALUE = null;
+    private static RegistrationService registrationService;
+    private final User user = new User();
+
+    @BeforeEach
+    void setUp() {
+        registrationService = new RegistrationServiceImpl();
```
RedArtelerist, line 26: I think it's better to initialize it in @BeforeAll method

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+        user.setAge(VALID_AGE);
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.PEOPLE.clear();
+    }
+
+    @Test
+    void register_userIsValid_Ok() throws RegistrationException {
```
RedArtelerist, line 38: Why do you need throws unchecked exception? We're not supposed to do that and anyway, in which method do you throw it?
```suggestion
    void register_userIsValid_Ok() {
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+    void register_ageIsUnderEighteen_NotOk() {
+        user.setAge(YOUNGER_AGE);
+        assertThrows(RegistrationException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void register_ageIsNegative_NotOk() {
+        user.setAge(NEGATIVE_AGE);
+        assertThrows(RegistrationException.class, () -> registrationService.register(user));
+    }
```
RedArtelerist, line 82: You can combine this two methods with invalid age

RedArtelerist (changes requested):
Good solution. try to improve it. Also you can check on empty strings in login and password tests, for example.

Andriy-dz (approved):
Very good code!

PR #725, opened 2023-09-04
Hm validation registration

don-bigdad:


src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final int VALID_AGE = 18;
+    private static RegistrationService registrationService;
+    private static Storage storage;
+    private static User validUser;
+
+    @BeforeAll
+    static void beforeAll() {
+        registrationService = new RegistrationServiceImpl();
+        storage = new Storage();
+        validUser = new User();
```
VolodymyrSkvira, line 30: 1: You don`t need to create a new Storage instance it is already imported in your class;
2: registrationService and User can both be initialised as class fields

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+        registrationService.register(validUser);
+    }
+
+    @Test
+    void sizeStorageWithTwoCorrectUsers_Ok() {
+        User validUser2 = new User();
+        validUser2.setAge(VALID_AGE);
+        validUser2.setLogin(ONE_MORE_LOGIN);
+        validUser2.setPassword(VALID_PASSWORD);
+        validUser2.setId((long) 13);
```
VolodymyrSkvira, line 54: You can simply change your @BeforeAll to @BeforeEach and set needed fields before every test automatically. In this case you`ll only need to set one value per test

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,63 @@
...
+    }
+
+    private User checkLengthOfData(User user) throws ValidationException {
+        if (user.getLogin().length() < MINIMAL_LENGTH) {
+            throw new ValidationException("Login length must be more than 6");
+        }
+        if (user.getPassword().length() < MINIMAL_LENGTH) {
+            throw new ValidationException("Password length must be more than 6");
+        }
+        return user;
```
VolodymyrSkvira, line 52: It would look more logicaly correct if you would group checks by login, pass and age and not by length. Anyway, I`ll let the mentor decide

VolodymyrSkvira (changes requested):
Good solution! However it could use some refactoring

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,62 @@
...
     public User register(User user) {
-        return null;
+        try {
+            checkUserDataNull(user);
+            checkLengthOfData(user);
+            checkIsUserAdult(user);
+            isLoginInStorage(user);
+        } catch (ValidationException e) {
+            throw new RuntimeException(e);
+        }
```
liliia-ponomarenko, line 23: Why do you need to catch an exception here?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,62 @@
...
+        return storageDao.add(user);
+    }
+
+    private User isLoginInStorage(User user) throws ValidationException {
```
liliia-ponomarenko, line 27:
```suggestion
    private User isLoginInStorage(User user)  {
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,62 @@
...
+        for (User userInStorage:Storage.PEOPLE) {
+            if (userInStorage.getLogin().equals(user.getLogin())) {
+                throw new ValidationException("User with this login name already exist");
+            }
+        }
```
liliia-ponomarenko, line 32: You can use method storageDao.get() instead  this logic

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,62 @@
...
+        return user;
+    }
+
+    private User checkIsUserAdult(User user) throws ValidationException {
```
liliia-ponomarenko, line 36:
```suggestion
    private User checkIsUserAdult(User user) {
```
remove it in all private methods

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,62 @@
...
+        if (user.getAge() < MINIMAL_AGE) {
+            throw new ValidationException("Age must equals or more than "
+                    + MINIMAL_AGE);
+        }
+        return user;
+    }
+
+    private User checkLengthOfData(User user) throws ValidationException {
+        if (user.getLogin().length() < MINIMAL_LENGTH) {
+            throw new ValidationException("Login length must be more than 6");
```
liliia-ponomarenko, line 46:
```suggestion
            throw new ValidationException("Login length must be more than " + MINIMAL_LENGTH);
```
use constant for messages too. if you change the constant value you need to change the message too (check it in other messages)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
liliia-ponomarenko: agree

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,97 @@
...
+    private static User validUser = new User();
+    private static User existUser = validUser;
+
+    @BeforeEach
+    void beforeEach() {
+        validUser.setAge(VALID_AGE);
+        validUser.setLogin(VALID_LOGIN);
+        validUser.setPassword(VALID_PASSWORD);
+        validUser.setId((long) 12);
+    }
```
liliia-ponomarenko, line 31: It is bad practice to use the same data for all tests. Better to create your own for each test. This improves readability and makes test cases more consistent

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Test
+    void userRegistration_Ok() {
+        registrationService.register(validUser);
+    }
+
+    @Test
+    void sizeStorageWithTwoCorrectUsers_Ok() {
+        User validUser2 = new User();
+        validUser2.setAge(VALID_AGE);
```
liliia-ponomarenko, line 46: better avoid using  names user1, user2

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,97 @@
...
+        validUser2.setLogin(ONE_MORE_LOGIN);
+        validUser2.setPassword(VALID_PASSWORD);
+        validUser2.setId((long) 13);
+        registrationService.register(validUser2);
+        registrationService.register(validUser);
+        assertEquals(2, Storage.PEOPLE.size());
+    }
+
+    @Test
+    void checkIsUserLoginLength_notOk() {
```
liliia-ponomarenko, line 56: names of test cases should give information on which parameter is invalid and why (check it everywhere)
```suggestion
    void invalidUserLoginLength_notOk() {
```

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,97 @@
...
+        User unvalidUserLogin = new User();
+        unvalidUserLogin.setAge(19);
+        unvalidUserLogin.setLogin(UNVALID_LOGIN);
+        unvalidUserLogin.setPassword(VALID_PASSWORD);
+        unvalidUserLogin.setId((long) 1);
```
liliia-ponomarenko, line 61: you don’t need to set Id before registration (usually it’s autoincrement value). fix it everywhere

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,97 @@
...
+import core.basesyntax.model.User;
+import core.basesyntax.service.RegistrationService;
+import core.basesyntax.service.RegistrationServiceImpl;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class RegistrationServiceImplTest {
+    private static final String VALID_LOGIN = "Username";
+    private static final String ONE_MORE_LOGIN = "Username2";
```
liliia-ponomarenko, line 16:
```suggestion
    private static final String ONE_MORE_LOGIN = "Username2";
```
bad naming, you can use FIRST_VALID_LOGIN, SECOND_VALID_LOGIN

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,97 @@
...
+    private static final String VALID_PASSWORD = "Password";
+    private static final String UNVALID_LOGIN = "12345";
```
liliia-ponomarenko, line 18:
```suggestion
    private static final String INVALID_LOGIN = "12345";
```
Let’s use invalid everywhere

liliia-ponomarenko (changes requested):
Good job! Let’s  fix a few mistakes ;)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
don-bigdad: > agree

I don't quite understand what you want me to do, we need to check the length of the password

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
+    private static final int MINIMAL_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        checkUserDataNull(user);
+        checkLengthOfData(user);
+        checkIsUserAdult(user);
+        isLoginInStorage(user);
```
NazarSmith228, line 18: these methods aren't placed in invocation order

I would expect that if `checkUserDataNull` is invoked first, then it should go first after the current method

it simplifies the reading process

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
+        return storageDao.add(user);
+    }
+
+    private User isLoginInStorage(User user) {
```
NazarSmith228, line 22: shouldn't this method be `void`?

and I would better call this method like `verifyLoginNotUsed`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new ValidationException("User with this login name already exist");
+        }
+        return user;
+    }
+
+    private User checkIsUserAdult(User user) {
```
NazarSmith228, line 29: shouldn't this method be `void`?

and I would better call this method like `verifyAdultUser`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
+        if (user.getLogin().length() < MINIMAL_LENGTH) {
+            throw new ValidationException("Login length must be more than " + MINIMAL_LENGTH);
+        }
+        if (user.getPassword().length() < MINIMAL_LENGTH) {
+            throw new ValidationException("Password length must be more than " + MINIMAL_LENGTH);
+        }
+        return user;
+    }
+
+    private User checkUserDataNull(User user) {
```
NazarSmith228, line 47: shouldn't this method be `void`?

and I would better call this method like `verifyCredentialsFilled`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
+
+    private User checkIsUserAdult(User user) {
+        if (user.getAge() < MINIMAL_AGE) {
+            throw new ValidationException("Age must equals or more than "
+                    + MINIMAL_AGE);
+        }
+        return user;
+    }
+
+    private User checkLengthOfData(User user) {
```
NazarSmith228, line 37: shouldn't this method be `void`?

and I would better call this method like `verifyCorrectCredentials`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new ValidationException("User with this login name already exist");
+        }
+        return user;
+    }
+
+    private User checkIsUserAdult(User user) {
+        if (user.getAge() < MINIMAL_AGE) {
+            throw new ValidationException("Age must equals or more than "
+                    + MINIMAL_AGE);
```
NazarSmith228, line 32: you can use `formatted` to avoid concatenation

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
+        }
+        return user;
+    }
+
+    private User checkLengthOfData(User user) {
+        if (user.getLogin().length() < MINIMAL_LENGTH) {
+            throw new ValidationException("Login length must be more than " + MINIMAL_LENGTH);
+        }
+        if (user.getPassword().length() < MINIMAL_LENGTH) {
+            throw new ValidationException("Password length must be more than " + MINIMAL_LENGTH);
```
NazarSmith228, line 42: you can use `formatted` to avoid concatenation

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class RegistrationServiceImplTest {
+    private static final String FIRST_VALID_LOGIN = "Username";
+    private static final String SECOND_VALID_LOGIN = "Username2";
+    private static final String VALID_PASSWORD = "Password";
+    private static final String INVALID_LOGIN = "12345";
+    private static final String INVALID_PASSWORD = "12345";
+    private static final int VALID_AGE = 18;
```
NazarSmith228, line 22: I consider constants in tests a bad pattern

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+    private User validUser = new User();
+    private User existUser = validUser;
+
+    @BeforeEach
+    void beforeEach() {
+        validUser.setAge(VALID_AGE);
+        validUser.setLogin(FIRST_VALID_LOGIN);
+        validUser.setPassword(VALID_PASSWORD);
+        validUser.setId(id++);
+    }
```
NazarSmith228, line 34: 1. you don't need to use ids here - they are usually filled within the DAO
2. you shouldn't have any generic users for all tests - instead create a new user with its data inside of each test method

due to that remark the only thing you need as a field is RegistrationService (which you're actually testing)

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @AfterEach
+    void tearDown() {
+        Storage.PEOPLE.clear();
+    }
+
```
NazarSmith228, line 40: general recommendation regarding structuring your tests - try to group them:
1. successful tests
2. login tests (empty, null, longer, shorter, used etc.)
3. password tests
4. age tests

this way it will be easier for reader to find a corresponding test scenario

NazarSmith228 (changes requested):

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,127 @@
...
+        Assertions.assertEquals(validUser,Storage.PEOPLE.get(0));
+    }
+
+    @Test
+    void checkUserIsNull_NotOk() {
+        assertThrows(RuntimeException.class, () -> registrationService.register(null));
+    }
+
+    @Test
+    void checkIsUserLoginLength_notOk() {
```
Andriy-dz, line 80:
```suggestion
    void register_shortLogin_notOk() {
```
the same with the names of other tests

Andriy-dz (approved):

PR #727, opened 2023-09-05
HW registration validation

OlehReex:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,42 @@
...
-        return null;
+        alreadyRegisteredValidator(user);
+        loginAndPasswordsValidator(user);
+        ageValidator(user);
+        return storageDao.add(user);
+    }
+
+    private void alreadyRegisteredValidator(User user) {
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidDataException("User with current login is already registered");
```
Parashchuk, line 23: Provide more informative error message

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+    private RegistrationService registrationService;
+    private User user;
+
+    @BeforeEach
+    void setUp() {
+        user = new User();
+        user.setId(111L);
+        user.setLogin("reexxx");
+        user.setPassword("qwerty");
+        user.setAge(25);
```
Parashchuk, line 23: Create constants for these values

Parashchuk (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
OlehReex: I'm not ssure it's good idea. But will see, what mentor say about this

Parashchuk (approved):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,42 @@
...
+    private static final int MIN_VALID_LENGTH = 6;
+    private static final int MINIMAL_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        alreadyRegisteredValidator(user);
+        loginAndPasswordsValidator(user);
+        ageValidator(user);
```
Elena-Bruyako, line 17: It's better to use verb in name of method. For example validateAge(); The same for other methods

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,42 @@
...
+
+    private void alreadyRegisteredValidator(User user) {
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidDataException("User with current login is already registered");
+        }
+    }
+
+    private void loginAndPasswordsValidator(User user) {
+        if (user.getLogin() == null || user.getLogin().length() < MIN_VALID_LENGTH) {
+            throw new InvalidDataException("Incorrect login. Length should be at least "
```
Elena-Bruyako, line 29: You could use String.format() to avoid concatenation

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,42 @@
...
+        }
+        if (user.getPassword() == null || user.getPassword().length() < MIN_VALID_LENGTH) {
+            throw new InvalidDataException("Incorrect password. Length should be at least "
+                    + MIN_VALID_LENGTH);
+        }
+    }
+
+    private void ageValidator(User user) {
+        if (user.getAge() == null || user.getAge() < MINIMAL_AGE) {
+            throw new InvalidDataException("Your age should been at least " + MINIMAL_AGE);
```
Elena-Bruyako, line 40: It's better to use this message: "User must be at least " + MINIMAL_AGE + " years old" with String.format()

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,42 @@
...
+        return storageDao.add(user);
+    }
+
+    private void alreadyRegisteredValidator(User user) {
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidDataException("User with current login is already registered");
+        }
+    }
+
+    private void loginAndPasswordsValidator(User user) {
```
Elena-Bruyako, line 27: It's better to create two separate methods for validate login and password.
Also, it's better to throw exception with message if login or password are null and throw exception when length aren't ok

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,42 @@
...
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_VALID_LENGTH = 6;
+    private static final int MINIMAL_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        alreadyRegisteredValidator(user);
```
Elena-Bruyako, line 15: validateUserRegistration or checkUserRegistration ?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.PEOPLE.clear();
+    }
+
+    @Test
+    void registerValidUser() {
```
Elena-Bruyako, line 33: let's stick to common mistakes about naming to testing method. 
Please, rename all methods according to the article

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private RegistrationService registrationService;
+    private User user;
+
+    @BeforeEach
+    void setUp() {
+        user = new User();
```
Elena-Bruyako, line 19: it's bad practice to create User here. 
Tests method should be self-explanatory so in such way each method has GIVEN block - initial parameters, WHEN block - which method you are going to test and THEN block - what happened after calling the method. 

So it's better to create User in every test method with different parameters for each case.

Elena-Bruyako (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,140 @@
...
+    void tearDown() {
+        Storage.PEOPLE.clear();
+    }
+
+    @Test
+    void register_validUser_ok() {
+        User actual = new User();
+        actual.setId(111L);
+        actual.setLogin("reexXx");
+        actual.setPassword("qwerty");
```
Elena-Bruyako, line 32: Who is actual and why we set it some parameters ?))))

Rename to user or currentUser or newUser or actualUser whatever.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @Test
+    void register_userWithValidAge() {
+        User actual = new User();
+        actual.setId(111L);
+        actual.setLogin("reexXx");
+        actual.setPassword("qwerty");
+        actual.setAge(25);
+        int expected = 25;
+        assertEquals(actual.getAge(), expected);
```
Elena-Bruyako, line 127: So you tested here that creating object in Java works well? )))) 

You need to register user and then check is age the same 

User actualUser = registrationService.register(user));
assertNotNull(actualUser);
assertNotNull(actualUser.getId());
assertEquals(user, actualUser);

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,140 @@
...
+        User actual = new User();
+        actual.setId(111L);
+        actual.setLogin("reexXx");
+        actual.setPassword("qwerty");
+        actual.setAge(-30);
+        assertThrows(InvalidDataException.class, () -> registrationService.register(actual));
+    }
+
+    @Test
+    void register_userWithValidAge() {
```
Elena-Bruyako, line 120: _ok in the end

Elena-Bruyako (changes requested):
Good job!

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
OlehReex: yeah, "actual" is not the best my idea)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
OlehReex: But I can't register actualUser with user logins.
To be honest, I don't understand why we need

assertNotNull(actualUser);
assertNotNull(actualUser.getId());
assertEquals(user, actualUser);

here

Elena-Bruyako: Discussed offline

Elena-Bruyako (approved):

PR #728, opened 2023-09-05
Implement register and test coverage

Parashchuk:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,44 @@
...
+                            + " symbols, but you have: "
+                            + user.getPassword().length());
+        }
+
+        if (user.getAge().compareTo(VALID_AGE) < 0) {
+            throw new RegisterException(
+                    "The age must have at least " + VALID_AGE
+                            + " years old, but yours is : "
+                            + user.getAge());
+        }
```
Vt0rtikV, line 42: I think you can split this ifs in different methods.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+    static void beforeAll() {
+        regService = new RegistrationServiceImpl();
+    }
+
+    @BeforeEach
+    void setUp() {
+        user = new User();
+        user.setLogin(VALID_LOGIN);
+        user.setAge(VALID_AGE);
+        user.setPassword(VALID_PASSWORD);
```
Vt0rtikV, line 31: You can make it in constructor. And i'm not sure about user in set(if you think that it's normal you can leave it for mentors, because i'm really not sure)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Test
+    void register_allParamsValid_ok() {
+        User actual = regService.register(user);
+        assertEquals(user, actual);
+    }
+
+    @Test
+    void register_nllUser_throwsException() {
```
Vt0rtikV, line 46: why not null?)

Vt0rtikV (changes requested):
Where is your user class at all?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
Parashchuk: I don't see the point to separate them, because its uses only once in code

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Parashchuk: There's no constructor tho, so i just used default setters and didn't change the initial class

src/main/java/core/basesyntax/service/RegisterException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+public class RegisterException extends RuntimeException {
```
Elena-Bruyako, line 3: It's better to rename to RegistrationException that describe that exception could be thrown during checking the parameters

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
Elena-Bruyako: Code duplication isn't the only reason for creating separate methods.
You method register should just register user and other methods could check or validate some parameters. Also, there are many if statements in one method.

In addition, it's better to check if login, password and age are null. 
In such case you will have more if statements and this will also cause separate methods to be used.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Elena-Bruyako: it's bad practice to create User here.
Tests method should be self-explanatory so in such way each method has GIVEN block - initial parameters, WHEN block - which method you are going to test and THEN block - what happened after calling the method.

So it's better to create User in every test method with different parameters for each case.


Also, it's bad to use constants in test. What if you have 100 valid login strings or password? In addition, if you have 50 test methods and more constant other developer should always stick to the beginning of test class to understand what exactly you are using for test. 


So, it's better to give all information what you are testing with particular parameters in each test methods.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Test
+    void register_allParamsValid_ok() {
+        User actual = regService.register(user);
+        assertEquals(user, actual);
+    }
+
+    @Test
+    void register_nullUser_throwsException() {
```
Elena-Bruyako, line 46: let's stick to the common mistake and rename methods like register_nullUser_notOk
The same for other methods

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Test
+    void register_nullPassword_throwsException() {
+        user.setPassword(null);
+        assertThrows(NullPointerException.class, () -> regService.register(user));
+    }
+
+    @Test
+    void register_invalidLogin_throwsException() {
+        user.setLogin("a");
```
Elena-Bruyako, line 71: What if login will be ""?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Test
+    void register_invalidAge_throwsException() {
+        user.setAge(12);
+        assertThrows(RegisterException.class, () -> regService.register(user));
+    }
+
+    @Test
+    void register_invalidPassword_throwsException() {
```
Elena-Bruyako, line 82: What if password will be more than min length?
The same for age?

Elena-Bruyako (changes requested):
Good job. Let's improve your solution!

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Parashchuk: In code there's totally valid case with all the conditions you described

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,55 @@
...
 public class RegistrationServiceImpl implements RegistrationService {
+    public static final int VALID_LOGIN_LENGTH = 6;
+    public static final int VALID_PASSWORD_LENGTH = 6;
+    public static final int VALID_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        Objects.requireNonNull(user);
```
NazarSmith228, line 16: this will throw `NullPointerException` with some default message 

however I would expect you to throw your custom exception with your message

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,55 @@
...
+        checkUserAlreadyExist(user);
+        checkLoginValid(user);
```
NazarSmith228, line 18: this sequence is incorrect

ideally you should verify that login is valid before checking whether it is duplicated

```suggestion
        checkLoginValid(user);
        checkUserAlreadyExist(user);
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,55 @@
...
+        if (user.getPassword().length() < VALID_PASSWORD_LENGTH) {
+            throw new RegistrationServiceException(
+                    "The password must have at least " + VALID_PASSWORD_LENGTH
+                            + " symbols, but you have: "
+                            + user.getPassword().length());
+        }
+    }
+
+    public void checkAgeAllowed(User user) {
+        if (user.getAge().compareTo(VALID_AGE) < 0) {
```
NazarSmith228, line 50: no need to use `compareTo` intead of simple `>` or `<`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,55 @@
...
+        if (user.getLogin().length() < VALID_LOGIN_LENGTH) {
+            throw new RegistrationServiceException(
+                    "The login must have at least " + VALID_LOGIN_LENGTH
+                            + " symbols, but you have: "
+                            + user.getLogin().length());
+        }
+    }
+
+    public void checkPasswordValid(User user) {
+        if (user.getPassword().length() < VALID_PASSWORD_LENGTH) {
```
NazarSmith228, line 41: what will happen if login or password will be blank like `      `

you should cover that as well

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,111 @@
...
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static RegistrationServiceImpl regService;
+    private User user;
```
NazarSmith228, line 15: you don't need any generic user for all tests - you should create users exclusively per test

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @BeforeAll
+    static void beforeAll() {
+        regService = new RegistrationServiceImpl();
+    }
+
+    @BeforeEach
+    void setUp() {
+        user = new User();
+    }
```
NazarSmith228, line 25: this method can be removed due to my comment above

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,111 @@
...
+        user.setPassword("1234");
+        assertThrows(RegistrationServiceException.class, () -> regService.register(user));
+    }
+
+    @Test
+    void register_userAlreadyExist_notOk() {
+        user.setLogin("validLogin");
+        user.setAge(18);
+        user.setPassword("12345678");
+        regService.register(user);
```
NazarSmith228, line 108: you should not call `register` to simulate existing user

instead you should use `Storage` directly

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,111 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
```
NazarSmith228, line 13: normal test name would be `RegistrationServiceTest` according to interface name - since you don't have multiple implementations

however this is a minor remark - you decide whether fix it or not :)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,111 @@
...
+    private static RegistrationServiceImpl regService;
```
NazarSmith228, line 14:
```suggestion
    private static RegistrationServiceImpl registrationService;
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,111 @@
...
+    @BeforeEach
+    void setUp() {
+        user = new User();
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.PEOPLE.clear();
+    }
+
```
NazarSmith228, line 31: general recommendation - group your tests by aim:
1. tests for login
2. tests for password
3. tests for age
etc.

this will simplify things for readers of your code

NazarSmith228 (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
Parashchuk: i used it only because i don't want to make check for null, is it not valid ?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Andriy-dz: not fixed

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,57 @@
...
+    public static final int VALID_LOGIN_LENGTH = 6;
+    public static final int VALID_PASSWORD_LENGTH = 6;
+    public static final int VALID_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new NullPointerException();
```
Andriy-dz, line 16: why NPE?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,112 @@
...
+        assertThrows(NullPointerException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void register_nullPassword_notOk() {
+        User user = new User();
+        user.setLogin("validLogin");
+        user.setAge(18);
+        user.setPassword(null);
+        assertThrows(NullPointerException.class, () -> registrationService.register(user));
```
Andriy-dz, line 64: let's catch not an NPE error, but a custom one. Same everywhere

Andriy-dz (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+public class RegistrationServiceException extends RuntimeException {
```
Elena-Bruyako, line 3: It's better RegistrationException

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,70 @@
...
+
+    public void checkLoginValid(User user) {
+        if (user.getLogin() == null) {
+            throw new RegistrationServiceException("The login can't be null");
+        }
+
+        int loginLength = user.getLogin().trim().length();
+        if (loginLength < VALID_LOGIN_LENGTH) {
+            throw new RegistrationServiceException(
+                    "The login must have at least " + VALID_LOGIN_LENGTH
```
Elena-Bruyako, line 42: You can use String.format() to avoid concatenation

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Elena-Bruyako: Not fixed

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,112 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static RegistrationServiceImpl registrationService;
```
Elena-Bruyako, line 13:
```suggestion
    private static RegistrationService registrationService;
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,112 @@
...
+    void register_emptyLogin_notOk() {
+        User user = new User();
+        user.setLogin("");
+        user.setAge(18);
+        user.setPassword("123456790");
+        assertThrows(RegistrationServiceException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void register_invalidAge_notOk() {
```
Elena-Bruyako, line 86: Let's add one case when age is negative

Like in common mistakes [Add tests for all possible User’s parameters (null login/password/age, user under 18/18/over 18 years old, negative age, and so on…).](https://mate-academy.github.io/jv-program-common-mistakes/java-core/junit/registration-validation)

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/service/RegistrationService.java
```diff
@@ -1,7 +1,71 @@
...
+    }
+
+    public void checkUserAlreadyExist(User user) {
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new RegistrationException(String.format(
+                    "User with %s as a login already exist", user.getLogin()));
+        }
+    }
+
+    public void checkLoginValid(User user) {
```
liliia-ponomarenko, line 34: this and the next methods should be private

liliia-ponomarenko (approved):
Good job ;)

PR #730, opened 2023-09-05
implement registration method and tested it

anigilator479:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,43 @@
...
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_LENGTH = 6;
+    private static final int MIN_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
-    public User register(User user) {
-        return null;
+    public User register(User user) throws UserInvalidDataException {
```
evgenkolner, line 14: I think it is not necessary

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,43 @@
...
+        if (Objects.isNull(user)) {
```
evgenkolner, line 15: fo me there are many if. you can create separate methods for every variable

src/test/java/core/basesyntax/service/UserInvalidDataException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+public class UserInvalidDataException extends Exception {
+    public UserInvalidDataException(String message) {
+        super(message);
+    }
+}
```
evgenkolner, line 7: for me good job

evgenkolner (approved):

src/main/java/core/basesyntax/service/RegistrationService.java
```diff
@@ -3,5 +3,5 @@
 import core.basesyntax.model.User;
 
 public interface RegistrationService {
-    User register(User user);
+    User register(User user) throws UserInvalidDataException;
```
NazarSmith228, line 6: why do we need this?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
NazarSmith228: agree

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,51 @@
...
+    private static final int MIN_LENGTH = 6;
+    private static final int MIN_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
-    public User register(User user) {
-        return null;
+    public User register(User user) throws UserInvalidDataException {
+        nullChecking(user);
+        userMeetsCriteriaCheck(user);
```
NazarSmith228, line 16: bad method naming

also I would suggest creating separate methods for different parts like:
1. verifyLogin(user)
2. verifyPassword(user)
3. verifyAge(user)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
NazarSmith228: agree, and also do not use `Objects.isNull` instead of plain old `== null`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,51 @@
...
+        }
+        if (user.getAge() == null) {
+            throw new UserInvalidDataException("User age can't be null");
+        }
+        if (user.getPassword() == null) {
+            throw new UserInvalidDataException("User password can't be null");
+        }
+    }
+
+    private void userMeetsCriteriaCheck(User user) throws UserInvalidDataException {
```
NazarSmith228, line 35: this method does too much, consider creating separate methods due to my comment above

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,51 @@
...
+            throw new UserInvalidDataException("User password length must be at least 6 symbols"
+                    + ", but length was:  " + user.getPassword().length());
+        }
+        if (user.getLogin().length() < MIN_LENGTH) {
+            throw new UserInvalidDataException("User login length must be at least 6 symbols,"
+                    + " but length was:  " + user.getLogin().length());
+        }
+        if (user.getAge() < MIN_AGE) {
+            throw new UserInvalidDataException("User age must be at least 18, but was: "
+                    + user.getAge());
```
NazarSmith228, line 50: you can use `formatted` method instead of concatenation for each exception message

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static final String EMPTY_STRING = "";
+    private static final String VALID_LOGIN = "Alexandr";
+    private static final String INVALID_LOGIN = "Alex";
+    private static final String VALID_PASSWORD = "Alex12345";
+    private static final String INVALID_PASSWORD = "12345";
+    private static final int VALID_AGE = 20;
+    private static final int INVALID_AGE = 17;
```
NazarSmith228, line 18: I consider constants in tests a bad pattern

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+    private User user;
+
+    @BeforeEach
+    void setUp() {
+        user = new User();
+        user.setPassword(VALID_PASSWORD);
+        user.setLogin(VALID_LOGIN);
+        user.setAge(VALID_AGE);
+        registrationService = new RegistrationServiceImpl();
+    }
```
NazarSmith228, line 29: you don't need any generic user for all tests - you should create users exclusively per test
```suggestion
    @BeforeEach
    void setUp() {
        registrationService = new RegistrationServiceImpl();
    }

```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @AfterEach
+    void tearDown() {
+        Storage.PEOPLE.clear();
+    }
+
+    @Test
+    void validRegistration_Ok() throws UserInvalidDataException {
```
NazarSmith228, line 37: you don't need `throws` in method signature - pls remove it

**Applicable for all tests**

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+        registrationService.register(user);
```
NazarSmith228, line 38: you don't do any assertions here - what's the point of this test?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Test
+    void duplicatesLogin_NotOk() throws UserInvalidDataException {
+        registrationService.register(user);
+        assertThrows(UserInvalidDataException.class, () -> registrationService.register(user));
```
NazarSmith228, line 44: you shouldn't call `register` twice to simulate duplicated login

you should prepare the already existing user directly via `Storage`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+        user.setLogin(VALID_LOGIN);
+        user.setAge(VALID_AGE);
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.PEOPLE.clear();
+    }
+
```
NazarSmith228, line 35: general recommendation - group your tests by aim:
1. tests for login
2. tests for password
3. tests for age
etc.

this will simplify things for readers of your code

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
```
NazarSmith228, line 11: normal test name would be `RegistrationServiceTest` according to interface name - since you don't have multiple implementations

however this is a minor remark - you decide whether fix it or not :)

NazarSmith228 (changes requested):

src/main/java/core/basesyntax/service/RegistrationService.java
anigilator479: it doesnt work if i remove it

liliia-ponomarenko: Let’s  extend  from RuntimeException

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,61 @@
...
+
+    private void verifyLogin(User user) throws UserInvalidDataException {
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new UserInvalidDataException(String.format("User with this login already "
+                    + "exists: %s", user.getLogin()));
+        }
+
+        if (user.getLogin().length() < MIN_LENGTH) {
+            throw new UserInvalidDataException(String.format("User login length must be at least "
+                    + "6 symbols, but length was: %d", user.getLogin().length()));
```
liliia-ponomarenko, line 45: use constant instead of 6 (check it in other places)

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,125 @@
...
+        Storage.PEOPLE.remove(user);
+    }
+
+    @Test
+    void duplicatesLogin_NotOk() {
+        user = new User();
+        user.setPassword("joker2004");
+        user.setLogin("same_login");
+        user.setAge(28);
+        assertThrows(core.basesyntax.exceptions.UserInvalidDataException.class,
```
liliia-ponomarenko, line 37:
```suggestion
        assertThrows(UserInvalidDataException.class,
```
Can you avoid using a path?

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,125 @@
...
+        Storage.PEOPLE.add(sameUser);
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.PEOPLE.remove(user);
+    }
+
+    @Test
+    void duplicatesLogin_NotOk() {
```
liliia-ponomarenko, line 32:
```suggestion
    void register_duplicatesLogin_NotOk() {
```
Better add method name. Because, if you add a new method it will be difficult to understand which method is testing

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,125 @@
...
+import core.basesyntax.db.Storage;
+import core.basesyntax.exceptions.UserInvalidDataException;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceTest {
+    private static RegistrationService registrationService;
+    private User user;
```
liliia-ponomarenko, line 14:
```suggestion
```
Let’s  create separate users for each  test case

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,125 @@
...
+        assertThrows(core.basesyntax.exceptions.UserInvalidDataException.class,
+                () -> registrationService.register(user));
+    }
+
+    @Test
+    void youngAge_NotOk() {
+        user = new User();
+        user.setPassword("Avadacedabra");
+        user.setLogin("Harry_Potter");
+        user.setAge(15);
```
liliia-ponomarenko, line 66:
```suggestion
        user.setAge(17);
```
it is best practice to use thresholds for testing. age 18/17, login 6/5

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
liliia-ponomarenko: not fixed

liliia-ponomarenko (changes requested):
Let’s  fix a few mistakes  ;)

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,120 @@
...
+    static void setUp() {
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.PEOPLE.clear();
+    }
+
+    @DisplayName("Duplicate login registration test")
```
Elena-Bruyako, line 26: Well done! Respect for using such annotation!

Elena-Bruyako (approved):

PR #737, opened 2023-09-06
done registration validation

TetianaDenchyk:


validatorr (approved):
LGTM!

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,54 @@
...
+    private void validateLogin(String login) {
+        if (login == null) {
+            throw new RegistrationException("Login can't be null.");
+        }
+        if (login.length() < MIN_LENGTH) {
+            throw new RegistrationException("Your login should contain "
+                    + MIN_LENGTH + " or more symbols.");
+        }
+        if (storageDao.get(login) != null) {
+            throw new RegistrationException("Your login: " + login + " already exists.");
```
Elena-Bruyako, line 33: Like an option 
"User with login already exists."

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,106 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.exception.RegistrationException;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
```
Elena-Bruyako, line 13: normal test name would be RegistrationServiceTest according to interface name - since you don't have multiple implementations

Minor comment

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,106 @@
...
+    void register_validUser_ok() {
+        User expected = new User("kateryna", "katya789", 19);
+        User actual = registrationService.register(expected);
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    void register_passwordIsNull_notOk() {
+        User user = new User("taylor12", null, 24);
+        for (User member : Storage.PEOPLE) {
```
Elena-Bruyako, line 76: Do we need this for loop ?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,54 @@
...
+        validateAge(user.getAge());
+        return storageDao.add(user);
+    }
+
+    private void validateLogin(String login) {
+        if (login == null) {
+            throw new RegistrationException("Login can't be null.");
+        }
+        if (login.length() < MIN_LENGTH) {
+            throw new RegistrationException("Your login should contain "
```
Elena-Bruyako, line 29: You can avoid concatenation by using String.format() method

Elena-Bruyako (changes requested):
Well done! 
Let's fix some small comments.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,58 @@
...
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new RegistrationException("The user doesn't exist");
+        }
+        validateLogin(user.getLogin());
+        validatePassword(user.getPassword());
+        validateAge(user.getAge());
```
NazarSmith228, line 20: it is better to pass `user` instance to each method so that it is responsibility of a method to check what's necessary

caller should not bother about it
```suggestion
        validateLogin(user);
        validatePassword(user);
        validateAge(user);
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,58 @@
...
+        return storageDao.add(user);
+    }
+
+    private void validateLogin(String login) {
+        if (login == null) {
+            throw new RegistrationException("Login can't be null.");
+        }
+        if (login.length() < MIN_LENGTH) {
```
NazarSmith228, line 28: what if login will be a blank line? you should also cover that

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,58 @@
...
+            throw new RegistrationException(String
+                    .format("Your login should contain %s MIN_LENGTH or more symbols.",
+                            MIN_LENGTH));
```
NazarSmith228, line 31: minor 

you can use `formatted` method directly on a string instead of static `format`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,58 @@
...
+            throw new RegistrationException(String.format("User with login %s already exists.",
+                    login));
+        }
+    }
+
+    private void validatePassword(String password) {
+        if (password == null) {
+            throw new RegistrationException("Your password can't be null.");
+        }
+        if (password.length() < MIN_LENGTH) {
```
NazarSmith228, line 43: what if login will be a blank line? you should also cover that

NazarSmith228 (changes requested):
Good job! Let's fix some final remarks

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
TetianaDenchyk: Could you explain why it is better please? 
When I pass user I should in the whole method use for example user.getLogin() while I could use just login instead

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
TetianaDenchyk: Thank you, very userful!

Elena-Bruyako (approved):

PR #739, opened 2023-09-06
Added solution JUnit

AlexRosenkreuz:
Implemented register method and created tests accordingly to the task

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -5,10 +5,32 @@ import core.basesyntax.dao.StorageDaoImpl;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user.getLogin() == null) {
+            throw new InvalidUserException("Login can't be null");
+        }
+        if (user.getPassword() == null) {
+            throw new InvalidUserException("Password can't be null");
+        }
+
+        if (user.getAge() < MIN_AGE) {
+            throw new InvalidUserException("Not valid age: "
+                    + user.getAge()
+                    + ". Min allowed age is " + MIN_AGE);
+        }
+
+        if (user.getLogin().length() < 6 || user.getPassword().length() < 6) {
+            throw new InvalidUserException("Login and password must be at least 6 characters long");
+        }
+
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidUserException("User with this login already exists");
+        }
+
+        return storageDao.add(user);
     }
 }
--- a/src/test/java/core/basesyntax/RegistrationServiceTest.java
+++ b/src/test/java/core/basesyntax/RegistrationServiceTest.java
@@ -1,0 +1,133 @@
+package core.basesyntax;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import core.basesyntax.model.User;
+import core.basesyntax.service.InvalidUserException;
+import core.basesyntax.service.RegistrationServiceImpl;
+import org.junit.jupiter.api.Test;
+
+public class RegistrationServiceTest {
+    private static final RegistrationServiceImpl REG_SERVICE = new RegistrationServiceImpl();
+    private static final int MIN_AGE = 18;
+
+    @Test
+    public void register_validUser_successfulRegistration() {
+        User user = new User();
+        user.setLogin("username");
+        user.setPassword("password123");
+        user.setAge(20);
+
+        User registeredUser = REG_SERVICE.register(user);
+        assertNotNull(registeredUser);
+        assertEquals(user.getLogin(), registeredUser.getLogin());
+    }
+
+    @Test
+    public void register_existingUser_throwException() {
+        User user1 = new User();
+        user1.setLogin("username");
+        user1.setPassword("password123");
+        user1.setAge(20);
+
+        User user2 = new User();
+        user2.setLogin("username");
+        user2.setPassword("anotherpassword");
+        user2.setAge(25);
+
+        Exception exception = assertThrows(InvalidUserException.class, () -> {
+            REG_SERVICE.register(user1);
+            REG_SERVICE.register(user2);
+        });
+
+        String expectedMessage = "User with this login already exists";
+        String actualMessage = exception.getMessage();
+
+        assertTrue(actualMessage.contains(expectedMessage));
+    }
+
+    @Test
+    public void register_shortLogin_throwException() {
+        User user = new User();
+        user.setLogin("short");
+        user.setPassword("password123");
+        user.setAge(20);
+
+        Exception exception = assertThrows(InvalidUserException.class,
+                () -> REG_SERVICE.register(user));
+
+        String expectedMessage = "Login and password must be at least 6 characters long";
+        String actualMessage = exception.getMessage();
+
+        assertTrue(actualMessage.contains(expectedMessage));
+    }
+
+    @Test
+    public void register_shortPassword_throwException() {
+        User user = new User();
+        user.setLogin("username");
+        user.setPassword("short");
+        user.setAge(20);
+
+        Exception exception = assertThrows(InvalidUserException.class,
+                () -> REG_SERVICE.register(user));
+
+        String expectedMessage = "Login and password must be at least 6 characters long";
+        String actualMessage = exception.getMessage();
+
+        assertTrue(actualMessage.contains(expectedMessage));
+    }
+
+    @Test
+    public void register_incorrectAge_throwException() {
+        User user = new User();
+        user.setLogin("username");
+        user.setPassword("password123");
+        user.setAge(17);
+
+        Exception exception = assertThrows(InvalidUserException.class,
+                () -> REG_SERVICE.register(user));
+
+        String expectedMessage = "Not valid age: "
+                + user.getAge()
+                + ". Min allowed age is " + MIN_AGE;
+        String actualMessage = exception.getMessage();
+
+        assertTrue(actualMessage.contains(expectedMessage));
+    }
+
+    @Test
+    public void register_nullLogin_throwException() {
+        User user = new User();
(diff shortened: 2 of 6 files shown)


yevhenkukharchuk (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
yevhenkukharchuk, line 1: I suggest separate parameter checks into separate methods and add one check with User=null and age(as it is Integer)=null

yevhenkukharchuk (changes requested):

Elena-Bruyako (changes requested):
Good job! 
Left some comments.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,47 @@
...
+    private static final int MIN_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+
+        validateUserParametersNotNull(user);
+        checkParameters(user);
+        checkExistingUser(user);
```
NazarSmith228, line 17: I would create separate method for each verification
```suggestion
        validateLogin(user); // includes check whether such login is used
        validatePassword(user);
        validateAge(user);
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,47 @@
...
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_AGE = 18;
+    private static final int MIN_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+
```
NazarSmith228, line 14: redundant empty line

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,47 @@
...
+        validateUserParametersNotNull(user);
+        checkParameters(user);
+        checkExistingUser(user);
+
```
NazarSmith228, line 18: redundant empty line

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,47 @@
...
+    }
+
+    private void checkParameters(User user) {
+        if (user.getAge() < MIN_AGE) {
+            throw new InvalidDataException("Not valid age: "
+                    + user.getAge()
+                    + ". Min allowed age is " + MIN_AGE);
+        }
+
+        if (user.getLogin().length() < MIN_LENGTH || user.getPassword().length() < MIN_LENGTH) {
```
NazarSmith228, line 41: you should also check whether it is not blank

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,155 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import core.basesyntax.model.User;
+import core.basesyntax.service.InvalidDataException;
+import core.basesyntax.service.RegistrationServiceImpl;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class RegistrationServiceTest {
+    private static final int MIN_AGE = 18;
```
NazarSmith228, line 15: this constant is redundant

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+        User user2 = new User();
+        user2.setLogin("username");
+        user2.setPassword("anotherpassword");
+        user2.setAge(25);
+
+        Exception exception = assertThrows(InvalidDataException.class, () -> {
+            registrationService.register(user1);
+            registrationService.register(user2);
+        });
```
NazarSmith228, line 50: bad setup

you should not call `register` twice to test duplicate login

instead you should perform all the necessary preparations before invoking the method under test

you can store user1 beforehand via `Storage.PEOPLE`

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,155 @@
...
+        user.setPassword("password123");
+        user.setAge(20);
+
+        Exception exception = assertThrows(InvalidDataException.class,
+                () -> registrationService.register(user));
+
+        String expectedMessage = "Login and password must be at least 6 characters long";
+        String actualMessage = exception.getMessage();
+
+        assertTrue(actualMessage.contains(expectedMessage));
```
NazarSmith228, line 71: `assertEquals("message", exception.getMessage())`

**apply to all places where necessary**

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,155 @@
...
+        user.setLogin("username");
+        user.setPassword("password123");
+        user.setAge(17);
+
+        Exception exception = assertThrows(InvalidDataException.class,
+                () -> registrationService.register(user));
+
+        String expectedMessage = "Not valid age: "
+                + user.getAge()
+                + ". Min allowed age is " + MIN_AGE;
```
NazarSmith228, line 102: there is no sense to do some concatenation - you should already know which message to expect

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+public class RegistrationServiceTest {
+    private static final int MIN_AGE = 18;
+    private RegistrationServiceImpl registrationService;
+
+    @BeforeEach
+    void setup() {
+        registrationService = new RegistrationServiceImpl();
+    }
+
```
NazarSmith228, line 22: seems you forgot to clear the state between tests

think of `@AfterEach` and what should be inside that method

NazarSmith228 (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
AlexRosenkreuz: Why do I need to check whether it's not blank, if in this case it has length of 0, which if less than MIN_LENGTH anyway? Or did I get the wrong idea?

NazarSmith228: @AlexRosenkreuz you got it wrong

empty string = `""`
blank string = `"         "` - as you can see it might have a valid length, but it is still invalid

there are 2 methods: `isEmpty` and `isBlank`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,47 @@
...
     public User register(User user) {
-        return null;
+        validateLogin(user);
+        validatePassword(user);
+        validateAge(user);
+        return storageDao.add(user);
+    }
+
+    private void validateLogin(User user) {
+        if (user.getLogin() == null) {
```
NazarSmith228, line 21: you can extract login to a separate variable to shorten the code below

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,47 @@
...
+            throw new InvalidDataException("Login can't be null");
+        }
+        if (user.getLogin().length() < MIN_LENGTH || user.getLogin().equals("")) {
```
NazarSmith228, line 24:
```suggestion
        if (user.getLogin().length() < MIN_LENGTH || user.getLogin().isBlank()) {
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,47 @@
...
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidDataException("User with this login already exists");
+        }
+    }
+
+    private void validatePassword(User user) {
+        if (user.getPassword() == null) {
+            throw new InvalidDataException("Password can't be null");
+        }
+        if (user.getPassword().length() < MIN_LENGTH) {
```
NazarSmith228, line 36: don't forget to check for blank

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
AlexRosenkreuz: Oh thank you. Gotta implement this

NazarSmith228 (changes requested):
Good job, let's fix some final remarks

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,50 @@
...
+        return storageDao.add(user);
+    }
+
+    private void validateLogin(User user) {
+        String login = user.getLogin();
+        if (login == null) {
+            throw new InvalidDataException("Login can't be null");
+        }
+        if (login.length() < MIN_LENGTH || login.isBlank()) {
+            throw new InvalidDataException("Login must be at least 6 characters long");
```
liliia-ponomarenko, line 26: better use constant 6 in the message. If you change the value you can forget to change the message

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,50 @@
...
+        }
+    }
+
+    private void validatePassword(User user) {
+        String password = user.getPassword();
+        if (password == null) {
+            throw new InvalidDataException("Password can't be null");
+        }
+        if (password.length() < MIN_LENGTH || password.isBlank()) {
+            throw new InvalidDataException("Password must be at least 6 characters long");
```
liliia-ponomarenko, line 39: the same here

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,172 @@
...
+    @Test
+    public void register_existingUser_throwException() {
+        User user1 = new User();
+        user1.setLogin("username");
+        user1.setPassword("password123");
+        user1.setAge(20);
+
+        registrationService.register(user1);
```
liliia-ponomarenko, line 47: better not to use register()  for preparation testing data when you test this method. You can add users to storage directly.

liliia-ponomarenko (changes requested):
a few minor mistakes ;)

liliia-ponomarenko (approved):
Great 😊

PR #747, opened 2023-09-09
tests implemented

katy-mozharovska:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,55 @@
...
-        return null;
+        validateUser(user);
+        return storageDao.add(user);
+    }
+
+    private void validateUser(User user) {
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidUserException("User with login '"
+                    + user.getLogin() + "' already exists.");
+        }
```
okuzan, line 24: put this after two next checks

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,55 @@
...
+
+        if (user.getLogin() == null) {
+            throw new InvalidUserException("Login is required");
+        }
+
+        if (user.getLogin().length() < 6) {
```
okuzan, line 30: use your constant

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,55 @@
...
+            throw new InvalidUserException("Invalid password '" + user.getPassword()
+                    + "'. Password must be at least '"
+                    + MIN_PASSWORD_LENGTH + "' characters long.");
+        }
+
+        if (user.getAge() == null) {
+            throw new InvalidUserException("Age is required");
+        }
+
+        if (user.getAge() < 18 || user.getAge() > 2147483647) {
```
okuzan, line 50: use your constant, and better omit max limit check
image

```suggestion
        if (user.getAge() < MIN_AGE) {
```

okuzan (changes requested):
You need only one positive test, where every value is valid and everything passes ok, you can keep one with multiple registrations if you wish. Short login, password - make sure to test precisely the edge case (6), and the value right below it (5), for the age - 18 and 17 (you have these). Also, it's more common to use lowercase after _underscore. Your private utility method is a constructor, so use a constructor instead. You can have a default valid user as a field, set it up in before each, and amend necessary fields for each test (set to short password for short password test. other values remain valid)

okuzan (approved):

PR #748, opened 2023-09-10
jv_HW

VasylPalokha:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Test
+    void register_validUser_notOk() {
+        User actualUser = registrationService.register(expectedUser);
+        assertEquals(expectedUser, actualUser);
+        assertNotNull(actualUser.getId());
+    }
+
+    @Test
+    void register_userIsPersoent_notOk() {
```
okuzan, line 102:
```suggestion
    void register_userIsPresent_notOk() {

```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,108 @@
...
+        registrationService.register(expectedUser);
+        assertThrows(InvalidDataException.class, () -> {
+            registrationService.register(expectedUser);
+        });
```
okuzan, line 106: we cannot use the same method we test in this scenario, add via dao directly
image

okuzan (changes requested):
Good job on the first try! Add _ok test where everything passes, assert it's added to storage and the returned user is correct

aaananas (approved):

PR #755, opened 2023-09-15
first

AndriiMarushchak:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,31 @@
...
+        this.storageDao = storageDao;
+    }
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null || user.getLogin() == null
+                || user.getPassword() == null
+                || user.getAge() == null) {
+            throw new RegistrationException("Invalid data");
```
okuzan, line 19: better to customize exception messages, it would be hard to pinpoint what went wrong with such a generic one

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,31 @@
...
+        }
+        if (user.getLogin().length() < 6 || user.getPassword().length() < 6
+                ||
+                user.getAge().intValue() < 18) {
```
okuzan, line 23: 6 and 18 should be constants

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+    private StorageDao storageDao;
+
+    @BeforeEach
+    void setup() {
+        storageDao = new StorageDaoImpl();
+        registrationService = new RegistrationServiceImpl(storageDao);
+    }
+
+    @Test
+    void register_validUser_Ok() {
```
okuzan, line 23: use lowercase after underscore

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+        User validUser = new User("validLogin", "validPassword", 18);
+        registrationService.register(validUser);
```
okuzan, line 25: here you tested only one thing - no exception is thrown upon valid user registration. It's hardly enough. If the method is empty and void, it will pass the test :) 
image

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Test
+    void register_invalidAge_NotOk() {
+        User invalidAgeUser = new User("invalidAge", "validPassword", 17);
+        org.junit.jupiter.api.Assertions.assertThrows(RegistrationException.class, ()
```
okuzan, line 31: add static import, don't write full name

src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Test
+    void register_nullPassword_NotOk() {
+        User nullPasswordUser = new User("validLogin", null, 18);
+        org.junit.jupiter.api.Assertions.assertThrows(RegistrationException.class, ()
+                -> registrationService.register(nullPasswordUser));
+    }
+
+    @Test
+    void register_nullAge_NotOk() {
```
okuzan, line 64: 0 isn't null

okuzan (changes requested):
Please go through the common mistakes checklist. Test edge case of length of exactly six.

PR #784, opened 2023-10-22
Jv registration validation

vadymhrnk:
added RegistrationService and covered it with tests

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -5,10 +5,33 @@ import core.basesyntax.dao.StorageDaoImpl;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MINIMUM_CHARACTERS = 6;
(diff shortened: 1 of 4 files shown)


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,33 @@
...
+        if (user.getPassword() == null || user.getPassword().length() < MINIMUM_CHARACTERS) {
+            throw new InvalidDataException("Password must be at least 6 characters.");
+        }
+
+        if (user.getAge() == null || user.getAge() < MINIMUM_AGE) {
+            throw new InvalidDataException("Age must be at least 18.");
+        }
+        storageDao.add(user);
+
+        return user;
```
TarasYurkovskyi, line 35: i think it would be better to create a separate helper method

TarasYurkovskyi (approved):
Good job)

kshuryhin (changes requested):
Nice try, but there are some points to resolve. Also, try to make literals and numbers in tests constant where it makes sense

kshuryhin (approved):

PR #786, opened 2023-10-22
Add unit-tests for the project

barbequeue:


src/main/java/core/basesyntax/dao/StorageDaoImpl.java
```diff
@@ -8,15 +8,22 @@ public class StorageDaoImpl implements StorageDao {
 
     @Override
     public User add(User user) {
-        user.setId(++index);
+        if (user == null) {
```
kozhukhovsky, line 11: no need to change this class, validation should be in the registration method

src/main/java/core/basesyntax/dao/StorageDaoImpl.java
```diff
@@ -8,15 +8,22 @@ public class StorageDaoImpl implements StorageDao {
...
+        user.setId(index++);
         Storage.people.add(user);
         return user;
     }
 
     @Override
     public User get(String login) {
         for (User user : Storage.people) {
-            if (user.getLogin().equals(login)) {
+            if (
```
kozhukhovsky, line 22: same

src/main/java/core/basesyntax/model/User.java
```diff
@@ -49,13 +49,14 @@ public boolean equals(Object o) {
             return false;
         }
         User user = (User) o;
-        return Objects.equals(login, user.login)
+        return Objects.equals(id, user.id)
```
kozhukhovsky, line 52: same

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,62 @@
...
+    public static final int MIN_PASSWORD_LENGTH = 6;
+    public static final int MIN_AGE = 18;
+
+    private static final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            return null;
```
kozhukhovsky, line 18: throw RegistrationException

src/test/java/core/basesyntax/dao/StorageDaoImplTest.java
```diff
@@ -0,0 +1,83 @@
...
+    public static void setupStorage() {
+        Storage.people.add(new User());
+        Storage.people.get(0).setLogin(null);
+        Storage.people.add(new User());
+        Storage.people.get(1).setLogin("admin");
+        Storage.people.add(null);
+    }
+
+    @Test
+    public void add_user_returnsSameUser() {
```
kozhukhovsky, line 28: image
It is necessary to rename all methods according to convention

src/test/java/core/basesyntax/dao/StorageDaoImplTest.java
```diff
@@ -0,0 +1,83 @@
...
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+public class StorageDaoImplTest {
+    private static final StorageDao storageDao = new StorageDaoImpl();
+
+    @BeforeAll
+    public static void setupStorage() {
```
kozhukhovsky, line 19: Each test should not depend on the results of the previous one. Thus, we will not be able to clean the storage, but we need to do this after each test.
Add to storage is only necessary in methods that check for the existence of users.

src/test/java/core/basesyntax/dao/StorageDaoImplTest.java
```diff
@@ -0,0 +1,83 @@
...
+        Storage.people.add(new User());
+        Storage.people.get(0).setLogin(null);
+        Storage.people.add(new User());
+        Storage.people.get(1).setLogin("admin");
+        Storage.people.add(null);
+    }
+
+    @Test
+    public void add_user_returnsSameUser() {
+        User user = new User();
```
kozhukhovsky, line 29: You can create a basic user with some fields and put it in a class variable. In each method, simply set the necessary fields for it, which will be tested in the method.

kozhukhovsky (changes requested):

src/test/java/core/basesyntax/dao/StorageDaoImplTest.java
barbequeue: I don't really understand the point about adding to storage being only necessary in specific methods. By the way, I understand, that each test method should be independant from the others, so i tried to achieve that in my current tests setting up the users **before** all the test methods

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
barbequeue: Shouldn't it be a `RuntimeException` instead? I think the absence of an user (`user == null`) is more like an issue with unproper handling a request to register an user. Like, i imagine an user pressing 'Sign Up' button without filling any required filelds - in that case the `User` class object with empty fields should be sent to `register` method, but not null. In that case an user gets an error - 'Login must be present' but not 'User should be present'

src/main/java/core/basesyntax/model/User.java
barbequeue: I don't think that `new User(0, 'admin', 'password', 20)` should be equal to `new User(1, 'admin', 'password', 20)` - they have different IDs. What do you think?

src/main/java/core/basesyntax/dao/StorageDaoImpl.java
barbequeue: Is this a requirement/restriction only for this task? Or more like a general rule of thumb?

I was thinking that responsibility of this unit is to put or retrieve a user from Storage, and in perspective this unit could be used later in other service with different kind of validation. So i think, this unit, should at least check the presence of a login field in `User` class object

src/test/java/core/basesyntax/model/UserTest.java
```diff
@@ -0,0 +1,150 @@
+package core.basesyntax.model;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+public class UserTest {
```
Ivan95kos, line 9: There is no need to test model classes
It is necessary to write unit tests for services
integration for controllers and repositories

Ivan95kos (approved):
Great job, with a minor comment

src/test/java/core/basesyntax/model/UserTest.java
barbequeue: Got it! Thank you ❤️

PR #788, opened 2023-10-23
First version of program was created. Added registration logic and tests for it

ponomvrenko:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,190 @@
...
+        user = new User();
+        user.setLogin("ponomvrenko");
+        user.setAge(21);
+        user.setPassword("qwerty1234");
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.people.clear();
+    }
```
nshtykh, line 40: You probably should move this method to the end.

nshtykh (changes requested):
Great job! I have only one minor comment

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,69 @@
...
+    private static final int MIN_LOGIN_LENGTH = 6;
+    private static final int MIN_PASSWORD_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (isValid(user)) {
+            storageDao.add(user);
+        }
```
kshuryhin, line 18: Great, but here I suggest to make validation methods `void` in order to remove redundant if statements

kshuryhin (changes requested):
LGTM, but also consider creating some constants for your unit tests, like `VALID_LOGIN`, `VALID_PASSWORD` etc.

kshuryhin (approved):

PR #792, opened 2023-10-23
Add implementation for the Registration Validation task

MinoDentori:
1. Add logic for the register method.
2. Create a test class.
3. Add a positive test, with all valid values.
4. Add the positive tests, with edge values.
5. Add the negative tests, with null values.
6. Add negative tests, with invalid, not null values.
7. Create the constants to store repeatable values.
8. Use BeforeAll and BeforeEach, to initialize fields.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,6 +31,11 @@
             <version>${junit-jupiter.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jacoco</groupId>
+            <artifactId>jacoco-maven-plugin</artifactId>
+            <version>0.8.11</version>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
@@ -48,7 +53,6 @@
                 </executions>
                 <configuration>
                     <configLocation>${maven.checkstyle.plugin.configLocation}</configLocation>
-                    <encoding>UTF-8</encoding>
                     <consoleOutput>true</consoleOutput>
                     <failsOnError>true</failsOnError>
                     <includeTestSourceDirectory>true</includeTestSourceDirectory>
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -2,13 +2,44 @@ package core.basesyntax.service;
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exceptions.RegistrationException;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int USER_MINIMAL_AGE = 18;
+    private static final int PASSWORD_LENGTH = 6;
+    private static final int LOGIN_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (isUserLoginAlreadyExist(user.getLogin())) {
+            throw new RegistrationException("User with this login already exist");
+        }
+        if (user.getAge() == null
+                || isMinimalUserAgeNotMet(user.getAge())) {
+            throw new RegistrationException("Minimal age requirement isn't met");
+        }
+        if (user.getLogin() == null
+                || isStringLengthNotMet(user.getLogin(), LOGIN_LENGTH)) {
+            throw new RegistrationException("Minimal login length requirement isn't met");
+        }
+        if (user.getPassword() == null
+                || isStringLengthNotMet(user.getPassword(), PASSWORD_LENGTH)) {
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,44 @@
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exceptions.RegistrationException;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int USER_MINIMAL_AGE = 18;
+    private static final int PASSWORD_LENGTH = 6;
+    private static final int LOGIN_LENGTH = 6;
```
DEz-1996, line 11: Will be better add "minimal" to constant names

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,44 @@
...
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            return null;
```
DEz-1996, line 17: Return null is bad practice
"You should create your custom unchecked exception and throw it in case of invalid data."

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @BeforeEach
+    void setUp() {
+        nullLoginPasswordUser.setAge(VALID_USER_AGE);
+        nullLoginPasswordUser.setLogin(null);
+        nullLoginPasswordUser.setPassword(null);
+        invalidAgeUser.setLogin(VALID_LOGIN);
+        invalidAgeUser.setPassword(VALID_PASSWORD);
+    }
```
DEz-1996, line 45: To simplify your code, you can initialize several default users in BeforeEach method and modify their fields with needed parameters in different tests

DEz-1996 (approved):
Good job, but please pay attention above my comments

pom.xml
```diff
@@ -31,6 +31,11 @@
             <version>${junit-jupiter.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jacoco</groupId>
+            <artifactId>jacoco-maven-plugin</artifactId>
+            <version>0.8.11</version>
+        </dependency>
```
kshuryhin, line 38: Just wondering, how much test coverage do you have in this task? :)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,47 @@
...
+        }
+        if (isMinimalUserAgeNotMet(user)) {
+            throw new RegistrationException("Minimal age requirement isn't met");
+        }
+        if (isStringLengthNotMet(user.getLogin(), MINIMAL_LOGIN_LENGTH)) {
+            throw new RegistrationException("Minimal login length requirement isn't met");
+        }
+        if (isStringLengthNotMet(user.getPassword(), MINIMAL_PASSWORD_LENGTH)) {
+            throw new RegistrationException("Minimal password length requirement isn't met");
+        }
```
kshuryhin, line 30: Cool, but what do you think about extracting entire validation stuff to the separate methods? Like this pseudocode
`void validate(user) throws Exception`

Thus we can just make a method call in the `register` and proceed next

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,143 @@
...
+        User actual = registrationService.register(defaultValidUser);
+        assertNotNull(actual);
+        assertEquals(defaultValidUser, actual);
+    }
+
+    @Test
+    void register_edgeLogin_ok() {
+        defaultValidUser.setLogin(EDGE_LOGIN);
+        User actual = registrationService.register(defaultValidUser);
+        assertNotNull(actual);
```
kshuryhin, line 75: Actually probably this null assertion is not needed, just because after that we check if `expected` is equal to `actual`

kshuryhin (changes requested):
Great job, just some minor suggestions from my side

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
MinoDentori: change structure, mb fixed )

pom.xml
MinoDentori: Got the reco for Jacoco using from Sofia, a little bit screwed up, during configuration, but for now, works okay.
After fixing your comments I got 70% from default Intelijj coverage and 91 from Jacoco.

kshuryhin (approved):

PR #795, opened 2023-10-23
implemented the registration method, custom exception and added the t…

annton-shakur:
…ests

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -1,14 +1,68 @@
 package core.basesyntax.service;
 
+import core.basesyntax.RegistrationException;
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
-    private final StorageDao storageDao = new StorageDaoImpl();
+    private static final String INVALID_PASSWORD_MESSAGE = "The provided password is invalid. "
+            + "It should include at least 6 characters!";
+    private static final String INVALID_LOGIN_MESSAGE = "The provided login is invalid. "
+            + "It should include at least 6 characters!";
+    private static final String LOGIN_EXISTS_MESSAGE = "The user with this login already exists.";
+    private static final String USER_IS_NULL_MESSAGE = "User cannot be null!";
+    private static final String AGE_INVALID_MESSAGE = "The age should be over 18.";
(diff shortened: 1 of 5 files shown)


AgroFix (approved):
Good job!!!

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,68 @@
...
+    private static final int MIN_LENGTH = 6;
+    private StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
+        if (checkIfUserIsNotNull(user)
+                && validateAge(user.getAge())
+                && validateLogin(user.getLogin())
+                && validatePassword(user.getPassword())
+                && checkIfLoginExists(user.getLogin())) {
```
kshuryhin, line 26: I believe we do not need our validation methods to have boolean return type, as it causes redundant if statement. Feel free to make them void, anyway if validation fails - there will be an exception

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+
+    private User testUser1 = new User();
+    private User testUser2 = new User();
+    private RegistrationService registrationService = new RegistrationServiceImpl();
+    private StorageDao storageDao = new StorageDaoImpl();
```
kshuryhin, line 20: It is more convenient to place resources initialisation to the `@BeforeAll `clause

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Test
+    void register_User_Is_Null_NotOk() {
+        testUser1 = null;
+        assertThrows(RegistrationException.class, () -> registrationService.register(testUser1));
+    }
+
+    @Test
+    void register_User_Login_Exists_NotOk() {
+        registrationService.register(testUser1);
```
kshuryhin, line 95: It would be better here to pre-sabe user with storage directly, as here we are testing registrationService and cannot fully rely on it's behaviour

kshuryhin (changes requested):
Good job, but there are some points to consider. Also, feel free to make some constant like `VALID_LOGIN`, `VALID_PASSWORD`, `INVALID_LENGTH_PASSWORD` in tests for readability improvement

PR #801, opened 2023-10-24
implement tests for register(User user) method

denysokhrymovych:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,46 @@
...
+        }
+        if (user.getLogin().length() < MINIMUM_LOGIN_LENGTH) {
+            throw new InvalidDataException("The login is shorter than 6 characters");
+        }
+        if (user.getPassword().length() < MINIMUM_PASSWORD_LENGTH) {
+            throw new InvalidDataException("The password is shorter than 6 characters");
+        }
+        if (user.getAge() < MINIMUM_AGE) {
+            throw new InvalidDataException("The age is less than 18");
+        }
```
MinoDentori, line 43: Cool, but what do you think about extracting the entire validation stuff to the separate methods? 
So we can have just like this pseudocode
```
        isUserNotNull(user);
        isUserLoginAlreadyExist(user.getLogin());
        validateUserAge(user.getAge());
        validateUserLogin(user.getLogin());
        validateUserPassword(user.getPassword());
```
Inside those methods, you can put your logic, mb combine null checking and value checking. It should increase the readability of your code and simplify further refactoring.
Thus we can just make a method call in the register and proceed next.

MinoDentori (changes requested):
Good job, no questions about the test, small recommendation about service implementation.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
mnyshenko: seems to be a valid point

mnyshenko (changes requested):
Critical points to consider

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,82 @@
...
+        checkIfUserIsNotNull(user);
+        checkIfIdIsNotNull(user);
+        checkIfLoginIsNotNull(user);
+        checkIfPasswordIsNotNull(user);
+        checkIfAgeIsNotNull(user);
+        checkIfUserExists(user);
+        checkLogin(user);
+        checkPassword(user);
+        checkAge(user);
+        Storage.people.add(user);
```
bhdnchui, line 26: Use userDao here.

bhdnchui (approved):
Good job!

PR #819, opened 2023-11-11
home-work (Registration validation)

ddryzhov:
implemented register method and wrote the tests

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -1,5 +1,6 @@
 package core.basesyntax.service;
 
+import core.basesyntax.InvalidUserDataException;
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
 import core.basesyntax.model.User;
@@ -9,6 +10,25 @@ public class RegistrationServiceImpl implements RegistrationService {
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new InvalidUserDataException("User cannot be null");
+        }
+
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidUserDataException("User already exists");
+        }
+
+        if (user.getLogin().length() >= 6 && user.getPassword().length() >= 6
+                && user.getAge() >= 18) {
+            try {
+                storageDao.add(user);
+                return user;
+            } catch (RuntimeException e) {
+                throw new InvalidUserDataException("Error during user registration: "
+                        + e.getMessage());
+            }
+        } else {
+            throw new InvalidUserDataException("Invalid user data");
+        }
     }
 }
--- a/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
+++ b/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
@@ -1,0 +1,95 @@
+package core.basesyntax.service;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.InvalidUserDataException;
+import core.basesyntax.db.Storage;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static RegistrationServiceImpl registrationService;
+
+    @BeforeAll
+    static void setUp() {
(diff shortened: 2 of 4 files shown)


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -9,6 +10,25 @@ public class RegistrationServiceImpl implements RegistrationService {
...
+            try {
+                storageDao.add(user);
+                return user;
+            } catch (RuntimeException e) {
+                throw new InvalidUserDataException("Error during user registration: "
+                        + e.getMessage());
+            }
+        } else {
+            throw new InvalidUserDataException("Invalid user data");
+        }
```
Ivan95kos, line 32: It is better to move the check to the private methods

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -9,6 +10,25 @@ public class RegistrationServiceImpl implements RegistrationService {
...
+        if (user == null) {
+            throw new InvalidUserDataException("User cannot be null");
+        }
+
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidUserDataException("User already exists");
+        }
+
+        if (user.getLogin().length() >= 6 && user.getPassword().length() >= 6
+                && user.getAge() >= 18) {
```
Ivan95kos, line 22: use constants instead of numbers

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
Ivan95kos, line 32: separate the if and try-catch

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+    static void setUp() {
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @Test
+    public void register_NullUser_ThrowsException() {
+        assertThrows(InvalidUserDataException.class, () -> {
+            registrationService.register(null);
+        });
+    }
```
Ivan95kos, line 26: fix the naming all methods
```suggestion
    public void register_nullUser_notOk() {
        assertThrows(InvalidUserDataException.class, () -> {
            registrationService.register(null);
        });
    }

```

Ivan95kos (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,46 @@
...
+    private void checkUserAlreadyExists(User user) {
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidUserDataException("User already exists");
+        }
+    }
+
+    private void validateUserData(User user) {
+        if (!(user.getLogin().length() >= LOGIN_LENGTH
+                && user.getPassword().length() >= PASSWORD_LENGTH && user.getAge() >= MIN_AGE)) {
+            throw new InvalidUserDataException("Invalid user data");
```
okuzan, line 43: Could be as is, but would be better to make informative error messages - if a password is too short, let's send according message, right now it would be hard to guess what exactly went wrong

okuzan (changes requested):
To avoid code duplication, you can setup one valid user and reuse it in your methods, resetting its fields as needed, However, your approach is viable too

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+        User existingUser = new User();
+        existingUser.setLogin("existingUser");
+        existingUser.setPassword("password123");
+        existingUser.setAge(20);
+        Storage.people.add(existingUser);
+
+        User newUser = new User();
+        newUser.setLogin("existingUser");
+        newUser.setPassword("password123");
+        newUser.setAge(20);
```
Sanderfuck, line 89: You can create template of valid user and reuse it any place where it`s needing and templates for not valid users

Sanderfuck (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,106 @@
...
+        return user;
+    }
+
+    private User createUserWithShortLogin() {
+        User user = new User();
+        user.setLogin("usrrr");
+        user.setPassword("password123");
+        user.setAge(20);
+        return user;
+    }
```
okuzan, line 53: you don't need those methods, just set validUser login too short for tooShortLogin test. You almost don't use your validUser

okuzan (changes requested):

okuzan (approved):
Well done!

PR #826, opened 2023-11-16
registration_validation_my_solution_firstCommit

OlegShvets1:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,69 @@
...
+    }
+
+    private void checkLogin(User user) {
+        if (user.getLogin() == null) {
+            throw new InvalidValidationException("User's login cannot be null");
+        }
+        if (user.getLogin().length() < VALID_LOGIN_LENGTH) {
+            throw new InvalidValidationException("Incorrect user`s login. Login must be at least "
+                    + VALID_LOGIN_LENGTH + " characters");
+        }
```
olekskov, line 42: Such differentiation of messages is usually not required.
Main goal is to give user clue what he should fix.
Also good practice is to add user's input into message so he can instantly get which value is wrong
```suggestion
        if (user.getLogin() == null || user.getLogin().length() < VALID_LOGIN_LENGTH) {
            String message = String.format("User's login is mandatory and should contain at least %s characters. Provided: '%s'",
                   VALID_LOGIN_LENGTH, user.getLogin)
            throw new InvalidValidationException(message);
        }
```

olekskov (approved):
Good job, approved.
Please check comment provided for further

PR #834, opened 2023-11-29
solution 29/11

VadimKadasiev:


VadimKadasiev:
> Java Cl checks failed. Before making a Pull Request (or a new commit to an already existing PR), you need to write the `mvn clean package` command in the terminal and make sure that all checks pass. After that, the mentor will be able to review the homework

Hi. I've checked my solution by "mvn clean package" and everything was OK, I can't get, what's wrong with my solution. I see in details, that maven can't find my exception class. But i've made it and pushed it exactly. Also i've tried to change it's destination.

PR #836, opened 2023-11-29
solution 29/11

VadimKadasiev:


olekskov (changes requested):
Task is not completed:
You should create your custom unchecked exception and throw it in case of invalid data.
Your task is to write as many tests as possible to cover all possible options.

Please feel free to ask in chat or attend Q&A session if you have any troubles with implementation

PR #854, opened 2023-12-25
Write tests, result-89%

slava-moiseev-kramatorsk:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Test
+    void invalid_login_test() {
+        assertThrows(RegistrationException.class,
+                () -> registrationService.register(invalidLoginUser));
+    }
+
+    @Test
+    void null_value_test() {
```
olekskov, line 41: Please recheck naming of your test methods.
According to task requirements:
For this task use such convention: ```<methodUnderTest>_<state>_<expectedBehavior>```; For example, if we are testing the method register with a null
user's age the test method name should be register_nullAge_notOk. notOk is because the test expects the register method to throw an exception.

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,36 @@
...
+        }
+        if (user.getLogin() == null || user.getLogin().isEmpty()) {
+            throw new RegistrationException("You forgot enter login, try again.");
+        }
+        if (user.getPassword() == null || user.getPassword().isEmpty()) {
+            throw new RegistrationException("You forgot enter password, try again.");
+        }
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new RegistrationException("This user already exists");
+        }
```
olekskov, line 36: please extract validation logic to separate private method to make public one short and clear

olekskov (changes requested):

olekskov (approved):
Good job

PR #859, opened 2024-01-01
added HW jv-registration-validation

arrem4eg:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @BeforeEach
+    void setUp() {
+        firstUser.setAge(MINIMUM_VALID_AGE);
+        firstUser.setPassword(SIX_CHARACTERS_STRING);
+        firstUser.setLogin(EIGHT_CHARACTERS_STRING);
+        secondUser.setAge(VALID_AGE);
+        secondUser.setPassword(EIGHT_CHARACTERS_STRING);
+        secondUser.setLogin(SIX_CHARACTERS_STRING);
```
okuzan, line 41: You may make changes to user class and create constructor if you want to

okuzan (changes requested):
Well done! Clean code, easy to read
Please use lowercase after underscore (like this _ok)
We need only one _ok test, where every field is valid, it's enough

PR #861, opened 2024-01-05
hm for review

vdbond:


src/main/java/core/basesyntax/exception/AuthenticationException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.exception;
+
+public class AuthenticationException extends Exception {
+
+    public AuthenticationException() {
+    }
+}
```
olekskov, line 7: 1. Runtime exception is more suitable here as we use it to interrupt application flow and do not plan to handle
2. unnecessary empty line
3. It's a bad practice to create exception without a message
```suggestion
public class AuthenticationException extends RuntimeException {
    public AuthenticationException(String message) {
        super(message);
    }
}
```

src/main/java/core/basesyntax/service/RegistrationService.java
```diff
@@ -1,7 +1,8 @@
 package core.basesyntax.service;
 
+import core.basesyntax.exception.AuthenticationException;
 import core.basesyntax.model.User;
 
 public interface RegistrationService {
-    User register(User user);
+    User register(User user) throws AuthenticationException;
```
olekskov, line 7: you do not need to modify interface, please rollback it to default state

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,57 @@
...
-        return null;
+    public User register(User user) throws AuthenticationException {
+        if (nullUserOrField(user)) {
+            throw new AuthenticationException();
+        }
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new AuthenticationException();
+        }
+        if (nonValidParameter(user)) {
+            throw new AuthenticationException();
```
olekskov, line 24: please add meaningful messages to your exceptions

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,161 @@
...
+    @Test
+    void justAddingFitUsers_Ok() {
+        try {
+            registrationService.register(new User("vlad222r", "afsgggwa", 25));
+            registrationService.register(new User("bob222", "fjjsafj", 18));
+            registrationService.register(new User("alice222", "viadkc", 65));
+        } catch (AuthenticationException e) {
+            fail("Can't add user");
+        }
+    }
```
olekskov, line 27: There is no need to explicitly handle exceptions in tests where it shouldn't be thrown.
If it's occured then test is automatically failed
```suggestion
    @Test
    void justAddingFitUsers_Ok() {
        registrationService.register(new User("vlad222r", "afsgggwa", 25));
        registrationService.register(new User("bob222", "fjjsafj", 18));
        registrationService.register(new User("alice222", "viadkc", 65));
    }
```

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @Test
+    void checkNullUser_notOk() {
+        try {
+            registrationService.register(null);
+        } catch (AuthenticationException e) {
+            return;
+        }
+        fail("User is not made");
```
olekskov, line 36: Please use assertThrows() method in tests instead of try/catch to check if proper exception occured.
Same issue for all other negative tests

olekskov (changes requested):

src/main/java/core/basesyntax/service/RegistrationService.java
Ivan95kos, line 7: not fixed
```suggestion
    User register(User user);
```

Ivan95kos (approved):
Great job, with a minor comment

PR #869, opened 2024-01-16
implemented test class  RegistrationServiceImplTest

petro2061:
Implemented test class testForRegistrationValidation. Tests describe different behavior that checs add not valid user.

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -2,13 +2,58 @@ package core.basesyntax.service;
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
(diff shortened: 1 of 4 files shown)


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,58 @@
...
+    private static final int MIN_LENGTH_OF_LOGIN = 6;
+    private static final int MIN_LENGTH_OF_PASSWORD = 6;
+    private static final int LOGIN_START_INDEX = 0;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
-    public User register(User user) {
-        return null;
+    public User register(User user) throws RegistrationServiceException {
+        if (user.getLogin() == null || user.getLogin().isEmpty()) {
```
kozhukhovsky, line 18: what if user is null?

kozhukhovsky (approved):
Good job!

PR #873, opened 2024-01-23
Registration-Validation

korets270:
Registration-Validation

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationService.java
+++ b/src/main/java/core/basesyntax/service/RegistrationService.java
@@ -1,7 +1,8 @@
 package core.basesyntax.service;
 
+import core.basesyntax.exceptions.ValidationException;
 import core.basesyntax.model.User;
 
 public interface RegistrationService {
-    User register(User user);
+    User register(User user) throws ValidationException;
 }
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -2,13 +2,49 @@ package core.basesyntax.service;
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exceptions.ValidationException;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MINIMAL_PASS_LENGTH = 6;
+    private static final int MINIMAL_LOGIN_LENGTH = 6;
+    private static final int MINIMAL_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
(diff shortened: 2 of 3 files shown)


okuzan (changes requested):
make sure your build passes before creating a PR - run `mvn clean verify` locally

src/main/java/core/basesyntax/exceptions/ExpectedException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.exceptions;
+
+public class ExpectedException extends Exception {
```
olekskov, line 3: 1) name of exception should represent case when it's used
2) please use extends RuntimeException in order to avoid interface signature changes
```suggestion
public class RegistrationException extends RuntimeException {
```

src/main/java/core/basesyntax/service/RegistrationService.java
```diff
@@ -1,7 +1,8 @@
 package core.basesyntax.service;
 
+import core.basesyntax.exceptions.ExpectedException;
 import core.basesyntax.model.User;
 
 public interface RegistrationService {
-    User register(User user);
+    User register(User user) throws ExpectedException;
```
olekskov, line 7: Please rollback interface to initial state

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,48 @@
...
+            throw new ExpectedException("Login must be longer than " + MINIMAL_LOGIN_LENGTH);
+        }
+        if (user.getAge() < MINIMAL_AGE) {
+            throw new ExpectedException("Age should be bigger than " + MINIMAL_AGE);
+        }
+        if (storageDao.get(user.getLogin()) == null) {
+            storageDao.add(user);
+            return user;
+        }
         return null;
```
olekskov, line 47: when user already exist you also should throw exception, not return null
```suggestion
        if (storageDao.get(user.getLogin()) != null) {
            throw new ...
        }
        return user;
```

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,146 @@
...
+        equalUser.setAge(18);
+        equalUser.setLogin("userForEquals");
+        equalUser.setPassword("password");
+        try {
+            registrationService.register(firstUser);
+            registrationService.register(equalUser);
+            assertEquals(1, Storage.people.size());
+        } catch (ExpectedException e) {
+            throw new RuntimeException("Cannot register such a User ", e);
+        }
```
olekskov, line 144: please use asserftThrows instead of try-catch

olekskov (changes requested):

kozhukhovsky (approved):

PR #883, opened 2024-02-01
added solution to the task

yanna-stepanova:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,61 @@
...
+    private static final int MIN_AGE = 18;
+    private static final int MAX_AGE = 150;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        checkForNull(user);
+        return (isValidValue(user)
+                && storageDao.get(user.getLogin()) == null) ? storageDao.add(user) : null;
```
olekskov, line 17: Returning null for registration is not a good idea. If email exists in DB we should also throw exception with proper message

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -3,12 +3,61 @@
...
+            }
+            if (user.getAge() == null) {
+                throw new RegistrationException("Age can't be null!");
+            }
+            return;
+        }
+        throw new RegistrationException("Object of user can't be null!");
+    }
+
+    public boolean isValidValue(User user) {
```
olekskov, line 36: Please use consistent approach.
Either all validations are void or boolean.
This will make understanding of code easier

olekskov (changes requested):

kozhukhovsky (approved):

PR #886, opened 2024-02-06
created unit tests

moskalenkosergii:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exception.ExpectedException;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_AGE = 18;
+    private static final int MIN_CHARACTERS = 6;
```
Vasul-Pankiv, line 10: better to divide it, cause requirements for password and login could be different

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        checkWrongValue(user);
```
Vasul-Pankiv, line 15: better name it "checkCredential" or "validateCredential"

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,55 @@
...
+        return storageDao.add(user);
+    }
+
+    private void checkWrongValue(User user) {
+
+        if (user == null) {
+            throw new ExpectedException("User can't be null");
+        }
```
Vasul-Pankiv, line 23: I think you can combine null and value validation (cause if Login is null it means it less than 6 character) same with password and age

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,210 @@
...
+        user.setLogin("userLogin");
+        user.setPassword("12345678");
+
+        newUser.setAge(32);
+        newUser.setLogin("userNewLogin");
+        newUser.setPassword("12345678New");
+    }
+
+    @Test
+    void register_userNull_notOk() {
```
Vasul-Pankiv, line 43: add some messages to notOk tests too 
like this "Test failed! Exception should be thrown if user is null!"

Vasul-Pankiv (changes requested):

olekskov (approved):

PR #894, opened 2024-02-19
written test and implementation of RegistrationServiceImpl

nastasiiaKulinich:


okuzan (changes requested):
Very nice job with your tests! Some improvements suggested

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Test
+    void register_validUserData_ok() {
+        User expectedUser = new User();
+        expectedUser.setLogin(VALID_LOGIN);
+        expectedUser.setPassword(VALID_PASSWORD);
+        expectedUser.setAge(VALID_AGE);
+        User actualUser = registrationService.register(expectedUser);
+        assertEquals(expectedUser, actualUser);
```
okuzan, line 130: let's also check that we can find the user by registered login
this test would pass if the register method would actually be

```
   @Override
        return user;
    }
  ```
  So we need to test it does add to the DB

okuzan (changes requested):

okuzan (approved):
Splendid 🔥 Comprehensive testing

PR #907, opened 2024-03-01
Fix

slizko1:


Bohadan-Markatov (approved):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,54 @@
...
+        passwordNullCheck(user);
+        ageCheck(user);
+        userRegistrationCheck(user);
+        loginLengthCheck(user);
+        passwordLengthCheck(user);
+        return storageDao.add(user);
+    }
+
+    private void passwordLengthCheck(User user) {
+        if (user.getPassword().length() < 6) {
```
karma-o, line 23: make 6 a constant

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
+package core.basesyntax.service;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.Test;
+
+public class RegistrationServiceImplTest {
```
karma-o, line 9: missing test cases:
* existing user with the same login(keep in mind the way how we should properly add the first user, check common mistakes for it)
* null user itself

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+    private static final int VALID_AGE = 20;
+    private static final int INVALID_AGE = 17;
+    private final RegistrationService registrationService = new RegistrationServiceImpl();
+    private final User user = new User();
+
+    @Test
+    public void registerValidUser_Ok() {
+        user.setLogin(VALID_LOGIN);
+        user.setPassword(VALID_PASSWORD);
+        user.setAge(VALID_AGE);
```
karma-o, line 23: let's not duplicate the `user` initialization, and bring it out to the `beforeEach`, where we should also reinitialize it with `new User()`, and let's rename this var to a `validUser` and it the test itself we can then leave just  single invalid field assignment

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.Test;
+
+public class RegistrationServiceImplTest {
+    private static final String VALID_LOGIN = "user123";
+    private static final String INVALID_LOGIN = null;
+    private static final String VALID_PASSWORD = "password123";
+    private static final String INVALID_PASSWORD = "nvdf";
+    private static final int VALID_AGE = 20;
```
karma-o, line 14: also better to test with edge cases
```suggestion
    private static final int VALID_AGE = 18;
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    public void registerUserWithNullLoginThrowRegistrationException_notOk() {
+        user.setLogin(INVALID_LOGIN); // Use the constant for null/empty login
+        user.setPassword(VALID_PASSWORD);
+        user.setAge(VALID_AGE);
+        RegistrationException exception = assertThrows(RegistrationException.class,
+                () -> registrationService.register(user));
+        assertEquals("Login can't be null", exception.getMessage()); // Check exception message
```
karma-o, line 35: let's do that for other tests as well

karma-o (changes requested):
Looks good so far, just some minor tings left

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Test
+    public void registerValidUser_Ok() {
+        User registeredUser = registrationService.register(validUser);
+        assertEquals(validUser, registeredUser);
+    }
+
+    @Test
+    void registerAlreadyAddedUserThrowRegistrationException_notOk() {
+        User registeredUser = registrationService.register(validUser);
```
karma-o, line 49: check common mistakes, we shouldn't use `register()` for adding the first user to the storage, because this method is under a test, and we can not guarantee that it is working properly, thus it can give us a false-positive result. Instead we should use `storage` or `storageDao`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,96 @@
...
+        RegistrationException exception = assertThrows(RegistrationException.class,
+                () -> registrationService.register(registeredUser));
+        assertEquals("User - " + registeredUser.getLogin()
+                + " is already registered", exception.getMessage());
```
karma-o, line 53: I'd rather extract this first parameter to some `expectedMessage` variable to improve readability

karma-o (approved):
Approved, but please check the comments regarding the common mistake

PR #910, opened 2024-03-03
Implemented the registration service and wrote unit tests for it.

ipantazi:
Created my custom unchecked exception.


okuzan (changes requested):
PR is empty, you forgot to commit your files
image

okuzan (approved):

PR #914, opened 2024-03-05
Registration implementation

makaroshyna:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+        user = null;
+        InvalidDataException invalidDataException = assertThrows(InvalidDataException.class,
+                () -> registrationService.register(user));
+        assertEquals(invalidDataException.getMessage(),
+                "User cannot be NULL!");
+    }
+
+    @Test
+    void register_loginAlreadyExists_notOk() {
+        registrationService.register(user);
```
karma-o, line 55: put the first user directly into `Storage` bypassing the register method, because we should not use a method that is under a test to arrange the testing data + storage is a more low-level way

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+        user.setLogin(null);
+        InvalidDataException invalidDataException = assertThrows(InvalidDataException.class,
+                () -> registrationService.register(user));
+        assertEquals(invalidDataException.getMessage(),
+                "Login should not be NULL!");
+    }
+
+    @Test
+    void register_loginLength_notOk() {
+        user.setLogin("login");
```
karma-o, line 74: add a constant with a name like `IVALID_LOGIN`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+        InvalidDataException invalidDataException = assertThrows(InvalidDataException.class,
+                () -> registrationService.register(user));
+        assertEquals(invalidDataException.getMessage(),
+                "The password must be at least " + MIN_PASSWORD_LENGTH
+                + " characters, but was " + user.getPassword().length() + "!");
+    }
+
+    @Test
+    void register_notValidUserAge_notOk() {
+        user.setAge(17);
```
karma-o, line 103: same here, and probably in some other places too, so please check for the values that could be constants

karma-o (changes requested):
Overall looking very good, just tweak some things here and there

PR #919, opened 2024-03-07
Added implementation for task registration validation.

nklimovych:
Added custom exception ValidationException & RegistrationServiceImplTest with all necessary test methods. Also implemented register method in RegistrationServiceImpl class.

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -2,13 +2,51 @@ package core.basesyntax.service;
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exception.ValidationException;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_LENGTH = 6;
+    private static final int MIN_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new ValidationException("User can't be null");
+        }
+
+        validateLogin(user.getLogin());
+        validatePassword(user.getPassword());
+        validateAge(user.getAge());
+
+        if (storageDao.get(user.getLogin()) == null) {
+            return storageDao.add(user);
+        }
+        throw new ValidationException("User already exist in database!");
+    }
+
+    private void validateLogin(String login) {
+        if (login == null) {
+            throw new ValidationException("Login can't be null");
+        } else if (login.length() < MIN_LENGTH) {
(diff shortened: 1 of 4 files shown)


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,51 @@
...
+        }
+        throw new ValidationException("User already exist in database!");
+    }
+
+    private void validateLogin(String login) {
+        if (login == null) {
+            throw new ValidationException("Login can't be null");
+        } else if (login.length() < MIN_LENGTH) {
+            throw new ValidationException("Minimum allowed login length is " + MIN_LENGTH);
+        }
```
vikkushnir, line 34: Not sure in this case, but Vadym once wrote that better to avoid "if else" construction
https://medium.com/swlh/return-early-pattern-3d18a41bba8
```suggestion
        if (login == null) {
            throw new ValidationException("Login can't be null");
        }
        if (login.length() < MIN_LENGTH) {
            throw new ValidationException("Minimum allowed login length is " + MIN_LENGTH);
        }
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,106 @@
...
+    void register_nullPassword_Not_Ok() {
+        user.setPassword(null);
+        String expected = "Password can't be null";
+        assertEquals(expected, assertThrowsException(user).getMessage());
+    }
+
+    private ValidationException assertThrowsException(User invalidUser) {
+        return assertThrows(ValidationException.class, () ->
+                registrationService.register(invalidUser));
+    }
```
vikkushnir, line 105: Remembered that Yevhenii approved such case on QnA :)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,106 @@
...
+        user.setLogin(INVALID_LOGIN);
+        String expected = "Minimum allowed login length is " + MIN_LENGTH;
+        assertEquals(expected, assertThrowsException(user).getMessage());
+    }
+
+    @Test
+    void register_invalidPassword_Not_Ok() {
+        user.setPassword(INVALID_PASSWORD);
+        String expected = "Minimum allowed password length is " + MIN_LENGTH;
+        assertEquals(expected, assertThrowsException(user).getMessage());
```
vikkushnir, line 78: Maybe would be better to check not only one variant
Screenshot 2024-03-07 at 22 30 11

vikkushnir (approved):
In total, looks fine to me, left some non critical comments

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
nklimovych: I read the article and have to agree - thanks for the good advice))

PR #920, opened 2024-03-07
jv-registration-validation solution

vikkushnir:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,64 @@
...
+    private static final int MIN_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        checkAge(user);
+        checkLogin(user);
+        checkPassword(user);
+        checkDuplicate(user);
```
EduarfF, line 18: I think that you need to create some method userRegister

EduarfF (approved):
Everything is good) Maybe you need to create static final var to your ages in Test, but I don't sure)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
vikkushnir: Not sure why I need it here this method

PR #923, opened 2024-03-08
added nessesary tests

AnyaRatusnyak:


rel1c-hub (approved):
Great job; I think you should use assertEquals method more in your tests)

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,65 @@
...
+                    + " already exists");
+        }
+        return true;
+    }
+
+    private boolean checkUserAge(Integer age) {
+        if (age == null) {
+            throw new RegistrationException("Age can't be null");
+        }
+        if (age < 18) {
```
karma-o, line 51: make 18 a constant

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Test
+    void register_nullUser_notOk() {
+        assertThrows(RegistrationException.class, () -> registrationService.register(null));
+    }
+
+    @Test
+    void register_nullLogin_NotOk() {
+        user.setLogin(null);
+        assertThrows(RegistrationException.class, () -> registrationService.register(user));
```
karma-o, line 62: let's also assert the error message that we get, same for other tests

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Test
+    void register_UserIsValid() {
+        user.setLogin(VALID_LOGIN);
+        user.setAge(VALID_AGE);
+        user.setPassword(VALID_PASSWORD);
+        assertDoesNotThrow(() -> registrationService.register(user),
+                "Registering user with valid data should not throw an exception");
+
```
karma-o, line 43: asserting that no exception has been thrown is fine, but the main thing that we need to check is that the user that we have registered actually got to the `Storage`
Also, redundant empty line

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Test
+    void register_OutputUserIsEqualtoInputUser_IsOk() {
+        User expected = user;
+        User actual = registrationService.register(user);
+        assertEquals(expected, actual, "Registered user should be equal to the original user");
```
karma-o, line 50: this assertion can just be a part of the above `register_UserIsValid()` test

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,105 @@
...
+        user.setLogin(EXIST_LOGIN);
+        user.setPassword(VALID_PASSWORD);
+        user.setAge(VALID_AGE);
+    }
+
+    @Test
+    void register_UserIsValid() {
+        user.setLogin(VALID_LOGIN);
+        user.setAge(VALID_AGE);
+        user.setPassword(VALID_PASSWORD);
```
karma-o, line 40: this is redundant because our user is already initialized with a valid data, btw we can rename this `user` var to `validUser` to have it clear

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,106 @@
...
+        assertDoesNotThrow(() -> registrationService.register(user),
+                "Registering user with valid data should not throw an exception");
+
+    }
+}
```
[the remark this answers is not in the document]
karma-o: in this tests we should arrenage the conditions before saving the user, by that I mean that we should put some user in the `Storage`(not using the `register()` method, check common mistakes) and after that try to register another user with the same login and assert an exception that we get from that.

BTW this logic that you have here now, should not throw any exception from the looks of it. If you want we can also investigate this part and try to figure out why it does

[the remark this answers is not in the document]
karma-o: btw, @rel1c-hub if you notice that something is wrong, next time try to explain why you think it is, and to fix that together first :wink:

karma-o (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+    void setUp() {
+        Storage.people.clear();
+        user = new User();
+        user.setLogin(EXIST_LOGIN);
+        user.setPassword(VALID_PASSWORD);
+        user.setAge(VALID_AGE);
+    }
+
+    @Test
+    void register_UserIsValid() {
```
karma-o, line 36:
```suggestion
    void register_UserIsValid_ok() {

```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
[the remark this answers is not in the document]
karma-o: this test is passing not because of the reasons we think it is, but because it has a null password and age, this is why need to assert the exception message - because it saves us from cases like that, when we think that it passes and all Is great, when in the reality it is written wrong and we wouldn't know if this functionality is broken.
Try to strictly follow my instructions that I have wrote in the previous comment about it,
here is another pseudo-code example:
```java
 @Test
    void register_userWithTheSameLogin_notOk() {
        // create some user or use the one that we already have
        // put it into storage via dao class or storage directly, because we cannot use register for it due to reason described above and in common mistakes
        // create another user wit the same login
        // register that second user
        // assert the resulting error
        // assert the message
}
```

karma-o (changes requested):
Now we just need to fix that one last test, do not be afraid to ask questions if you have any 😉

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Test
+    void register_ExistUser_notOk() {
+        Storage.people.add(user);
+        assertThrows(InvalidDataException.class,
+                () -> registrationService.register(user),
+                "Added the same user");
+    }
+
```
karma-o, line 114: string param in the `assertThrows()` is a custom message that we can add and receive when the test fails, needed to describe the reason why a test failed
We should also assert the exception message as I said earlier
```suggestion
        InvalidDataException ex = assertThrows(InvalidDataException.class,
                () -> registrationService.register(user),
                "we shouldn't allow to register a user with the same login as one of the existing users"); 
        // assert that ex.getMessage() is equal to some expected message
    }
```

karma-o (approved):
Approved! Just left 1 last comment, please check it

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
AnyaRatusnyak: Ok) I understand!! Thank you))

PR #924, opened 2024-03-08
jv-registration-validation-completed

margosalii:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,37 @@
...
+        if (user == null) {
+            throw new RegistrationException("User can't be null");
+        }
+        if (user.getLogin() == null) {
+            throw new RegistrationException("Login can't be null");
+        }
+        if (user.getPassword() == null) {
+            throw new RegistrationException("Password can't be null");
+        }
+        if (user.getLogin().equals(storageDao.get(user.getLogin()))) {
```
olekskov, line 23: This logic will not work if you have user with such email but request holds different password as they won't be equal.
You should just check existence of user with the requested email regardless of all other fields

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,147 @@
...
+    void tearDown() {
+        Storage.people.clear();
+    }
+
+    @Test
+    void noSuchUsersLoginInStorage_Ok() {
+        newUser.setLogin("abfhlfjsn");
+        actual = storageDao.get(newUser.getLogin());
+        assertNotEquals(newUser, actual);
+    }
```
olekskov, line 48: You should test registration service, not dao.
Positive case will just obtain new user for registration
Negative case should register user and then obtain second call with user with same login

olekskov (changes requested):

olekskov (approved):
Good job

PR #926, opened 2024-03-08
feat: initial commit

rel1c-hub:
first try on testing

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -1,14 +1,54 @@
 package core.basesyntax.service;
 
 import core.basesyntax.dao.StorageDao;
-import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exception.InvalidUserException;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
-    private final StorageDao storageDao = new StorageDaoImpl();
+    private static final int USER_MIN_AGE = 18;
+    private static final int USER_PASSWORD_MIN_LENGTH = 6;
+    private static final int USER_LOGIN_MIN_LENGTH = 6;
+
+    private final StorageDao storageDao;
+
+    public RegistrationServiceImpl(StorageDao storageDao) {
+        this.storageDao = storageDao;
+    }
 
     @Override
     public User register(User user) {
-        return null;
+        validateUserInput(user);
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidUserException("User with login '"
+                    + user.getLogin() + "' already exists");
+        }
+        return storageDao.add(user);
+    }
+
+    private void validateUserInput(User user) {
+        validateUserLogin(user.getLogin());
+        validateUserPassword(user.getPassword());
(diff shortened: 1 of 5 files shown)


TonyH277 (approved):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,53 @@
...
-        return null;
+        validateUserInput(user);
+        return storageDao.add(user);
+    }
+
+    private void validateUserInput(User user) {
+        validateUserExist(user);
+        validateUserLogin(user.getLogin());
+        validateUserPassword(user.getPassword());
+        validateUserAge(user.getAge());
```
karma-o, line 24: validate existence is an expensive operation, so lets' move it after the the simple ones
```suggestion
        validateUserLogin(user.getLogin());
        validateUserPassword(user.getPassword());
        validateUserAge(user.getAge());
        validateUserExist(user);
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,53 @@
...
+                    + user.getLogin() + "' already exists");
+        }
+    }
+
+    private void validateUserAge(Integer age) {
+        if (age == null || age < USER_MIN_AGE) {
+            throw new InvalidUserException("Users under "
+                    + USER_MIN_AGE + " are not allowed!");
+        }
+    }
```
karma-o, line 39: separate the null checks from the other ones, to have a clear message because right now if age is null you are saying it is under the min age, which is misguiding
same for other validation methods

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Test
+    void register_userWithNullAge_notOk() {
+        User user = createValidUser();
+        user.setAge(null);
+        InvalidUserException exception = assertThrows(InvalidUserException.class,
+                () -> registrationService.register(user));
+        assertEquals(exception.getMessage(), "Users under "
+                + USER_MIN_AGE + " are not allowed!");
```
karma-o, line 51: you can actually see that problem here, but you just covered it with tests that are expecting this strange behavior, we should fix business logic when you face something like this

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,119 @@
...
+                () -> registrationService.register(user));
+        assertEquals(exception.getMessage(), "Password must be at least "
+                + USER_LOGIN_PASSWORD_MIN_LENGTH + " characters");
+    }
+
+    @Test
+    void register_userWithExistingLogin_notOk() {
+        User user1 = createValidUser();
+        User user2 = createValidUser();
+        registrationService.register(user1);
```
karma-o, line 98: we should not put existing user via `service.register()` method, because this method is under the test right now, and we shouldn't use it to arrange the tests data, use low-level approach with `storageDao` or `Storage` instead
check common mistakes

karma-o (changes requested):

karma-o (approved):
GJ :fire:

PR #927, opened 2024-03-08
build success

MolchanovAlexander:


Bohadan-Markatov (approved):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,88 @@
...
+        existingUser.setAge(20);
+        existingUser.setLogin("existingUser");
+        existingUser.setPassword("password123");
+        storage.add(existingUser);
+    }
+
+    @BeforeEach
+    void setUp() {
+        registrationService = new RegistrationServiceImpl();
+    }
```
karma-o, line 32: * why do you need a new `registrationService` before each test?
*  your user that you are creating in each test should be a lass field to create and initialize it in the `BeforeEach` and called a `validUser`, then you just set a single field that he needs for specific tests
* Existing user on the other hand should be only in the duplicate login test, since it is used only there
* magic values should be constants

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Test
+    void register_UserOk_notNull() throws ValidationException {
+        User user = new User();
+        user.setAge(20);
+        user.setLogin("username");
+        user.setPassword("password123");
+        User registeredUser = registrationService.register(user);
+        assertNotNull(registeredUser);
+        assertEquals(user, registeredUser);
```
karma-o, line 42: you can and should also check that that user has actually got to the `Storage.people`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+import core.basesyntax.dao.StorageDao;
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exceptions.ValidationException;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
```
karma-o, line 15: we should have a tests for:
* null user
* null login/password/age
* business-wise invalid login/password/age
* existing user

make sure you have everything, because your coverage may look great percentage-wise, but that is mostly because you have 2 checks throwing the same exception and by writing a test for only one of those conditions your percentage grows but the actual and "real" test coverage stays not that great

karma-o (changes requested):

src/main/java/core/basesyntax/exceptions/ValidationException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.exceptions;
+
+public class ValidationException extends Exception {
```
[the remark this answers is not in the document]
MolchanovAlexander: because it could happen only in Runtime

karma-o (approved):
nice one

PR #929, opened 2024-03-08
jv-registration-validation task is done

bodya4243:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,22 @@
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exceptions.ValidationException;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int VALID_LOGIN_LENGTH = 2;
```
Nikname2303, line 9: why 2? we need user's login is at least 6 characters

Nikname2303 (approved):
Look good, only one minor comment)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+    private static final int INVALID_AGE = 16;
+    private static final int VALID_AGE = 18;
+    private static final long VALID_ID = 1234567890L;
+    private static final String INVALID_PASSWORD = null;
+    private static final String VALID_PASSWORD = "qwerty";
+    private static final String VALID_LOGIN = "nameEx";
+    private static final String INVALID_LOGIN = "name";
+
+    private final User user = new User();
+    private final RegistrationService registrationService = new RegistrationServiceImpl();
```
karma-o, line 24: * user should not be final, and should be initialized before each test to achieve a proper isolation
* proper order of the fields:
  * constants
  * services
  * test data

```suggestion
    private final RegistrationService registrationService = new RegistrationServiceImpl();
    
    private User user;
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.exceptions.ValidationException;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
```
karma-o, line 14: I don't see a case to check the existing user case

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+                () -> registrationService.register(user));
+        assertEquals("Validation error",
+                exception.getMessage(), "Incorrect validation error message");
+    }
+
+    @Test
+    public void register_sameLogin_notOk() {
+        assertFalse(Storage.people.contains(user),
+                "User with the same login should not exist");
+    }
```
karma-o, line 76: Right now you are just asserting that there is no user with such login in the storage, which is expected because there is no logic in putting it there, but it should test the the `register()` behavior when there is already a user in the storage, and we try to register another one with the same login
order of actions for this test: 
* put the user to the storage(without using the `register()` method, because it is under the test, see common mistakes for details)
* try to register a user with the same login
* assert the resulting exception and it's message

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+    @AfterEach
+    void tearDown() {
+        Storage.people.clear();
+    }
+
+    @Test
+    public void register_AddToStorage_Ok() {
+        User registeredUser = registrationService.register(user);
+        assertEquals(user, registeredUser, "The registered user should match the valid user");
+    }
```
karma-o, line 43: we can also assert if the user actually got to the `Storage` here

karma-o (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,40 @@
...
+        if (user.getLogin().length() < VALID_LOGIN_LENGTH) {
+            throw new ValidationException("Invalid login: must be at least 6 characters long.");
+        }
+
+        if (user.getPassword().length() < VALID_PASSWORD_LENGTH) {
+            throw new ValidationException("Invalid password: must be at least 6 characters long.");
+        }
+
+        if (user.getAge() < VALID_AGE) {
+            throw new ValidationException("Invalid age: must be at least 18 years old.");
```
karma-o, line 38: use constants that you already have in the exception messages instead of hardcoded values, to avoid changing them manually there in case they will change in the future

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,104 @@
...
+    private static final String INVALID_LOGIN = "name";
+
+    private User user;
+    private StorageDao storage;
+    private RegistrationService registrationService = new RegistrationServiceImpl();
+
+    @BeforeEach
+    void setUp() {
+        user = new User();
+        storage = new StorageDaoImpl();
```
karma-o, line 32: why do we need a new `storageDao` each time?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,104 @@
...
+        user.setAge(VALID_AGE);
+        user.setId(VALID_ID);
```
karma-o, line 34: user does not have an id assigned until he is registered
```suggestion
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,104 @@
...
+                exception.getMessage(), "Incorrect validation error message");
+    }
+
+    @Test
+    public void register_sameLogin_notOk() {
+        storage.add(user);
+        registrationService.register(user);
+        assertTrue(Storage.people.contains(user),
+                "User with the same login should not exist");
+    }
```
karma-o, line 103: This test does not make any sense, we assert that we have some user in the storage(there are actually 2 identical users in it), and if it does not - we fail the test with a message that it shouldn't exist,

This `register()` call should throw an exception which we should assert here because we can not allow `storage` to have 2 users with the same login as per our requirements. Please be more careful about the task description next time because this logic is missing in the service.

karma-o (changes requested):
I recommend you to debug the existing user test case and see how it works now before fixing the comments

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,67 @@
...
+    private void validateUser(User user) {
+        isLoginExists(user);
+        isLoginNull(user);
+        isPasswordNull(user);
+        isPasswordInvalid(user);
+        isLoginInvalid(user);
+        isAgeInvalid(user);
+    }
+
+    private void isLoginInvalid(User user) {
```
andrii-hoienko, line 30: Better to rename to 'validateLogin' or 'checkLogin', because word 'is' means that boolean variable will be returned

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,106 @@
...
+    void setUp() {
+        user = new User();
+        user.setAge(VALID_AGE);
+        user.setPassword(VALID_PASSWORD);
+        user.setLogin(VALID_LOGIN);
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.people.clear();
```
andrii-hoienko, line 37: This should be done in BeforeEach to make sure that db is clear before test execution. For example other tests can use storage as well, and if they don`t clean up, test that you wrote could be broken

andrii-hoienko (changes requested):
Final charge

andrii-hoienko (changes requested):
Final charge

PR #930, opened 2024-03-08
First solution

Nikname2303:


karma-o:
Missing peer review, you should ask your colleagues for it to not wait that long on the mentor review

src/main/java/core/basesyntax/InvalidDataException.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public class InvalidDataException extends Exception {
+    public InvalidDataException(String message) {
+        super(message);
+    }
+
```
karma-o, line 7:
```suggestion
```

src/main/java/core/basesyntax/InvalidDataException.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public class InvalidDataException extends Exception {
```
karma-o, line 3: should it be a checked exception?

src/main/java/core/basesyntax/dao/StorageDao.java
```diff
@@ -1,9 +1,10 @@
 package core.basesyntax.dao;
 
+import core.basesyntax.InvalidDataException;
 import core.basesyntax.model.User;
 
 public interface StorageDao {
-    User add(User user);
+    User add(User user) throws InvalidDataException;
```
karma-o, line 7:
```suggestion
    User add(User user);
```

src/main/java/core/basesyntax/dao/StorageDaoImpl.java
```diff
@@ -1,13 +1,21 @@
...
     @Override
-    public User add(User user) {
+    public User add(User user) throws InvalidDataException {
+        try {
+            if (Storage.people.contains(user)) {
+                throw new InvalidDataException("Ooops, your login already exist");
+            }
+        } catch (InvalidDataException e) {
+            throw new InvalidDataException("Ooops, your login already exist");
+        }
```
karma-o, line 18: you should not modify the `Storage` or `StorageDao` classes in this task

src/main/java/core/basesyntax/service/RegistrationService.java
```diff
@@ -1,7 +1,8 @@
 package core.basesyntax.service;
 
+import core.basesyntax.InvalidDataException;
 import core.basesyntax.model.User;
 
 public interface RegistrationService {
-    User register(User user);
+    User register(User user) throws InvalidDataException;
```
karma-o, line 7:
```suggestion
    User register(User user);
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,32 @@
...
+            throw new InvalidDataException("Oops, your input data is not valid");
+        }
+        if (user.getPassword().length() >= MINIMUM_CHARACTERS
+                && user.getLogin().length() >= MINIMUM_CHARACTERS
+                && user.getAge() >= MINIMUM_AGE) {
+            storageDao.add(user);
+            return user;
+        } else {
+            throw new InvalidDataException("Oops, your input data is not valid");
+        }
```
karma-o, line 30: this is not readable
* we should split it into methods that will be responsible for checking a single thing, like `validateLogin()` and call them one by one in `register()`
* we should not have any try/catch blocks here, RunTime exceptions like `NullPointerException` should be avoided and not handled by code when we already struck up in ot
* exception messages should be informative, which means it should contain information on what exactly went wrong, was it a null value? or invalid value for the business logic point of view, we also should indicate what value would be valid(putting min age to that message for example)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,79 @@
+package core.basesyntax.service;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.InvalidDataException;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    public User getDefaultValidUser() {
```
karma-o, line 10: why public?
this logic should also be called in the `beforeEach` method to reinitialize the `validUser` calss-field(which we also need to add) to not create this user every time in the tests themselves, but to just update a single field that we need it to have in the corresponding test.
this beforeEach should also clear the storage to have our tests isolated and not mix-up the test data between them

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+        User user = new User();
+        user.setLogin("validLogin");
+        user.setId(38628L);
+        user.setPassword("unuwjf8j2");
+        user.setAge(22);
+        return user;
+    }
+
+    @Test
+    void registrationDataIsValid() throws InvalidDataException {
```
karma-o, line 20: all tests should follow the same naming convention mentioned in the common mistakes:
`<methodUnderTheTest>_<parametersThatItRecieves>_<Ok/notOk>`
example: `register_nullLogin_notOk`
check out other point from it too

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+        RegistrationService registrationService = new RegistrationServiceImpl();
```
karma-o, line 21: do not create a new service each time make it a class-level variable, we have been doing it this way for several weeks now

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+        User user = getDefaultValidUser();
+        try {
+            registrationService.register(user);
+        } catch (InvalidDataException e) {
+            throw new InvalidDataException("Oops, your input data is not valid");
+        }
```
karma-o, line 27: why are we expecting an exception if this user is valid?
is the valid data test we should assert the result the we get from the `register()` method call and storage actually contains this user now

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Test
+    void ageIsNotOk() {
+        RegistrationService registrationService = new RegistrationServiceImpl();
+        User user = getDefaultValidUser();
+        user.setAge(16);
```
karma-o, line 34: all magic values should be constants, I suggest naming them like `VALID_AGE` or `INVALID_LOGIN`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+        assertThrows(InvalidDataException.class, () -> {
```
karma-o, line 35: we should also assert the message that we have in that exception, same for all other tests

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Test
+    void userAlreadyExist() {
+        RegistrationService registrationService = new RegistrationServiceImpl();
+        User user = getDefaultValidUser();
+        assertThrows(InvalidDataException.class, () -> {
+            registrationService.register(user);
+        });
+    }
```
karma-o, line 78: * The logic that you have here right now throws an exception only because your tests are not isolated enough and the users from other tests are staying in the `storage` after their completion. you need to clear it after/before each test.
* after you are done with the first point: You do not have a logic for putting a user that should already be present in the storage when you call the `register()` method. Hint: the first user should be put **not** using a `register()` but something more low-level because register is not guaranteed to work properly since it is under a test, check common mistakes for details on that

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,79 @@
+package core.basesyntax.service;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.InvalidDataException;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
```
karma-o, line 9: overall we should have the next tests here:
* one valid data case
* null user itself
* null case for each field
* invalid case for each field
* the case with the user with such login already existing

karma-o (changes requested):

src/main/java/core/basesyntax/dao/StorageDaoImpl.java
```diff
@@ -1,13 +1,14 @@
...
+import core.basesyntax.InvalidDataException;
 import core.basesyntax.db.Storage;
 import core.basesyntax.model.User;
 
 public class StorageDaoImpl implements StorageDao {
     private static Long index = 0L;
 
     @Override
-    public User add(User user) {
+    public User add(User user) throws InvalidDataException {
```
andrii-hoienko, line 11:
```suggestion
    public User add(User user) {
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,73 @@
...
+        if (user.getAge() == null) {
+            throw new InvalidDataException("Oops, age is not exist");
+        }
+        if (user.getAge() < MINIMUM_AGE) {
+            throw new InvalidDataException("Sorry, for registration you must be at least "
+                    + MINIMUM_AGE + " years");
+        }
+    }
+
+    private void isUserExist(User user) {
```
andrii-hoienko, line 62: Method which starts from 'is' means that boolean will be returned, but you have void, so I believe it should be renamed

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,73 @@
...
+        if (user.getLogin() == null) {
+            throw new InvalidDataException("Your login is not exist");
+        }
+        if (user.getLogin().length() < MINIMUM_CHARACTERS) {
+            throw new InvalidDataException("Your login is too short");
+        }
+    }
+
+    private void validatePassword(User user) {
+        if (user.getPassword() != null && user.getPassword().length() >= MINIMUM_CHARACTERS) {
```
andrii-hoienko, line 38: No reason to add 'happy check', please remove it

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,109 @@
...
+    private static final String VALID_PASSWORD = "unuwjf8j2";
+    private static final long VALID_ID = 38628L;
+    private static final int VALID_AGE = 22;
+    private static final String INVALID_LOGIN = "val";
+    private static final String INVALID_PASSWORD = "unu";
+    private static final int INVALID_AGE = 16;
+
+    private RegistrationService registrationService;
+    private User validUser;
+    private User nullUser;
```
andrii-hoienko, line 23: No reason for this variable, you can just pass null if it is required

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,109 @@
...
+    private static final String VALID_LOGIN = "validLogin";
+    private static final String VALID_PASSWORD = "unuwjf8j2";
+    private static final long VALID_ID = 38628L;
+    private static final int VALID_AGE = 22;
+    private static final String INVALID_LOGIN = "val";
+    private static final String INVALID_PASSWORD = "unu";
+    private static final int INVALID_AGE = 16;
+
+    private RegistrationService registrationService;
+    private User validUser;
```
andrii-hoienko, line 22:
```suggestion
    private User user; //This user is valid only in initial state, than is code cases you make him invalid and it is weird when you provide validUser and expect error
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
andrii-hoienko: not resolved

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,73 @@
 package core.basesyntax.service;
 
+import core.basesyntax.InvalidDataException;
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.db.Storage;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MINIMUM_CHARACTERS = 6;
```
andrii-hoienko, line 10:
```suggestion
    private static final int MIN_LENGTH= 6; //Or MINIMAL_LENGTH
```

andrii-hoienko (changes requested):
Nice, but some improvements required

src/main/java/core/basesyntax/dao/StorageDaoImpl.java
andrii-hoienko: Not resolved

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
andrii-hoienko: Not fully resolved

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,75 @@
...
+        }
+    }
+
+    private void validatePassword(User user) {
+        if (user.getPassword() == null) {
+            throw new InvalidDataException("Your password is not exist");
+        }
+        if (user.getPassword().length() >= MIN_LENGTH) {
+            return;
+        }
```
andrii-hoienko, line 44:
```suggestion
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,75 @@
...
+        return user;
+    }
+
+    private void validateLogin(User user) {
+        if (user.getLogin() == null) {
+            throw new InvalidDataException("Your login is not exist");
+        }
+        if (user.getLogin().length() >= MIN_LENGTH) {
+            return;
+        }
```
andrii-hoienko, line 31:
```suggestion
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,75 @@
...
+    }
+
+    private void validateAge(User user) {
+        if (user.getAge() == null) {
+            throw new InvalidDataException("Oops, age is not exist");
+        }
+        if (user.getAge() >= MINIMUM_AGE) {
+            return;
+        }
```
andrii-hoienko, line 57:
```suggestion
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,75 @@
...
 
+import core.basesyntax.InvalidDataException;
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.db.Storage;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_LENGTH = 6;
+    private static final int MINIMUM_AGE = 18;
```
andrii-hoienko, line 11: Please be consistent use word MIN or MINIMUN for both cases simultaneously

andrii-hoienko (changes requested):
Final charge

PR #932, opened 2024-03-08
Java core junit homework jv registrtion validator

azzilian:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,51 @@
...
+    private static final int MIN_LOGIN_LENGTH = 6;
+    private static final int MIN_PASSWORD_LENGTH = 6;
+    private static final int MIN_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (!checkLogin(user)) {
+            throw new UserRegistrationException("Invalid login");
```
JohnGrey17, line 17: maybe it will be better if your exception became more informative 
for example why login is Invalid?
```suggestion
            throw new UserRegistrationException("Invalid login. length of  should be greater 6 characters ");
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,167 @@
...
+    private static final String INVALID_EMPTY_PASSWORD = "";
+    private static final int INVALID_ZERO_AGE = 0;
+
+    private RegistrationServiceImpl registrationService;
+    private User user;
+
+    @BeforeEach
+    public void setUp() {
+        registrationService = new RegistrationServiceImpl();
+    }
```
JohnGrey17, line 36: i guess we should create new User here .
and registrationService = new RegistrationServiceImpl(); don't need to create before each

JohnGrey17 (commented):
Looks good added some comments

JohnGrey17 (approved):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
azzilian: The idea of User creation - each time I create user with new fields according to test goal so no need to create before each since I would need to setup its field anyway

I have class field
private RegistrationServiceImpl registrationService; - and I create new object for each test

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,56 @@
...
     @Override
     public User register(User user) {
-        return null;
+        boolean checkRegistrationAllowance = checkLogin(user.getLogin())
+                && checkPassword(user.getPassword())
+                && checkAge(user.getAge())
+                && !userExists(user);
+        if (!checkRegistrationAllowance) {
+            throw new UserRegistrationException("Registration failed ");
+        }
```
karma-o, line 23: Returning a boolean value from those validation methods is not a bad idea but building it into a single check looks hard to read and is actually redundant, this exception here will never be thrown because other ones will be thrown earlier, so I'd rather just call all of the `check...()` methods in order and that's it
BTW you also have no check if the `user` itself is null

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,56 @@
...
+        return storageDao.add(user);
+    }
+
+    private boolean checkPassword(String password) {
+        if (password == null || password.length() < MIN_PASSWORD_LENGTH) {
+            throw new UserRegistrationException("Invalid password , password shoudnt be less than "
+                    + MIN_PASSWORD_LENGTH + " characters");
+        }
```
karma-o, line 31: separate the null check from it, because right now if it is null you just say that user has to be of valid age, and nothing about null, which is misguiding.
Same for other validations

src/test/java/core/basesyntax/dao/StorageDaoImplTest.java
```diff
@@ -0,0 +1,47 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import core.basesyntax.model.User;
+import java.time.LocalDateTime;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class StorageDaoImplTest {
+    private static final String VALID_LOGIN = "test_login_created_at " + LocalDateTime.now();
```
karma-o, line 13: Does this `time` have a human-readable `toString()`? I think we need to use `DateTimeFormatter` for it to look nice, not sure tho, but please double-check

src/test/java/core/basesyntax/dao/StorageDaoImplTest.java
```diff
@@ -0,0 +1,49 @@
...
+        assertTrue(actual, "User id cannot be lower than 0");
+    }
+
+    @Test
+    public void search_UserLoginSavedCorrectly_Ok() {
+        User user = new User(VALID_LOGIN, VALID_PASSWORD, VALID_AGE);
+        storageDao.add(user);
+        String login = storageDao.get(VALID_LOGIN).getLogin();
+        assertEquals(user.getLogin(), login, "Login of saved user must "
+                + "be same as login used for registration");
```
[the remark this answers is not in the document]
karma-o: not fixed

src/test/java/core/basesyntax/dao/StorageDaoImplTest.java
```diff
@@ -0,0 +1,47 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import core.basesyntax.model.User;
+import java.time.LocalDateTime;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class StorageDaoImplTest {
```
karma-o, line 12: we do not really need to test storage in this task, but since you already did - good job! :+1:

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,176 @@
...
+        try {
+            registrationService.register(user);
+        } catch (UserRegistrationException e) {
+            return;
+        }
+        fail("Password cannot be null, exception about login check failed should see you");
+    }
+
+    @Test
+    public void register_NullPassExeption_notOk() {
```
karma-o, line 78: we should avoid shortened names and abbreviations
```suggestion
    public void register_NullPasswordException_notOk() {

```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,176 @@
...
+        user.setPassword(null);
+        try {
+            registrationService.register(user);
+        } catch (UserRegistrationException e) {
+            return;
+        }
+        fail("Password cannot be null, exception about password check failed should see you");
```
karma-o, line 85: Better to use `assertThrows()` instead

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    @Test
+    public void register_ZeroAge_notOk() {
+        user = new User(VALID_LOGIN, VALID_PASSWORD, INVALID_ZERO_AGE);
```
karma-o, line 90:
```suggestion
        user.setAge(INVALID_ZERO_AGE);
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+import core.basesyntax.db.Storage;
+import core.basesyntax.exeptions.UserRegistrationException;
+import core.basesyntax.model.User;
+import java.time.LocalDateTime;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class RegistrationServiceImplTest {
```
karma-o, line 15: I do not see a case of trying to register a user when one with the same login is already registered, let's add it
also check common mistakes for details on how to properly add the first one to the storage :wink:

karma-o (changes requested):

PR #935, opened 2024-03-09
Jv registration validation hw branch

JohnGrey17:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,52 @@
...
+        if (password == null || password.length() <= MINIMAL_PASSWORD_LENGTH) {
+            throw new PasswordValidateException("password must be longer than six characters");
+        }
+    }
+
+    private void ageValidation(int age) {
+        if (age < MINIMAL_AGE) {
+            throw new AgeValidationException(
+                    "User age must be at least " + MINIMAL_AGE);
+        }
```
azzilian, line 51: Not sure if we need 3 different Exceptions with same logic,  but i think ik ok since more easy to read

azzilian, line 51: Also as I remember its recommended not to use void methods, in lecture it was recommended to use at least boolean - like method better to return something

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+    private static final String DEFAULT_LOGIN = "curvabobr";
+    private static final String DEFAULT_PASSWORD = "Defaultpassword17";
+    private static final int MINIMAL_AGE = 18;
+    private static final int MINIMAL_PASSWORD_LENGTH = 6;
+    private static final int MINIMAL_LOGIN_LENGTH = 4;
+    private static final int FIRST_LOGIN_CHARACTER = 0;
+
+    private final RegistrationService registrationService = new RegistrationServiceImpl();
+    private final StorageDao storageDao = new StorageDaoImpl();
+    private User user;
```
azzilian, line 24: I have a question for mentor mostly - isnt this better to create separate Test class to test StorageDaoImpl separately?

azzilian (approved):
I think you can also add some methods with assertThrows and try\catch fail. But over all everything is ok for me

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
karma-o: we do not need to test storage in this task, overall dao classes are tested only if there is some custom logic, but in this case it is just a simulation, so no need to tests it

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
karma-o: overall - I agree, but here we either throw an exception - or return true for example, so I do not see a reason for that return

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,52 @@
...
+
+    private void checkExistUserInDataBase(String login) {
+        loginValidation(login);
+        if (storageDao.get(login) != null) {
+            throw new LoginValidateException("consumer with " + login + "login is already "
+                    + "exist please try another one login try to login with password");
+        }
+    }
+
+    private void passwordValidation(String password) {
```
karma-o, line 41: method names should be verbs, same for others
```suggestion
    private void validatePassword(String password) {
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+    void login_is_not_null_ok() {
+        String actual = user.getLogin();
+        assertNotNull(actual);
+    }
+
+    @Test
+    void password_is_not_null_ok() {
+        String actual = user.getPassword();
+        assertNotNull(actual);
+    }
```
karma-o, line 77: You are testing that your test data has not null value in some fields here. What we should do is set null to user password/login/age and try to register him asserting the resulting exception and it's message

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+        int actual = user.getLogin().length();
+        assertTrue(actual > MINIMAL_LOGIN_LENGTH, "login length should be longer than "
+                + MINIMAL_AGE);
+    }
+
+    @Test
+    void login_started_with_letter_ok() {
+        char actual = user.getLogin().charAt(FIRST_LOGIN_CHARACTER);
+        assertTrue(Character.isLetter(actual), "your login should start with any letter");
+    }
```
karma-o, line 65: all of those tests, should test the `register()` method of the service, and you are not even calling it there. We should verify the behavior of the `register()` with the different parameters, like short password/login, under the age, equal to the min age, and other edge cases.
Also, do not forget to tests the existing user validation

karma-o (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @AfterEach
+    void tearDown() {
+        Storage.people.clear();
+    }
+
+    @Test
+    void register_nullUser_throwsNullPointerException() {
+        assertThrows(NullPointerException.class, () -> registrationService.register(null));
```
karma-o, line 41: This test is testing that our service is written poor 😅 
When we throw or expect an NPE - it is wrong, in every way possible, what we should do is add a validation to the service and check if user is null - throw our custom exception with a message that we can not accept the null user and thus he is not valid. Here we should assert this exact custom exception and it's message

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Test
+    void register_invalidAge_throwsInvalidAgeException() {
+        User user = new User();
+        user.setLogin(VALID_LOGIN);
+        user.setAge(INVALID_AGE);
+        user.setPassword(VALID_PASSWORD);
```
karma-o, line 49: You have a valid user fully initialized in the class, use it instead, by just setting him invalid age for this test
Same for other tests

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Test
+    void register_shortPassword_throwsPasswordValidateException() {
+        User user = new User();
+        user.setLogin(VALID_LOGIN);
+        user.setAge(VALID_AGE);
+        user.setPassword(INVALID_PASSWORD);
+        assertThrows(PasswordValidateException.class,
+                () -> registrationService.register(user),
+                "password length should be minimum 6 symbols");
```
karma-o, line 62: This string param is not used for asserting that this exception has this message, it is used to show you that custom test failure information. so here for example it could be something like `in this test service should have thrown a PasswordValidateException, as a response to a short password, but it did not`
What we should do with the exception message and how can we assert it? Like that:
```java
PasswordValidateException thrownException = assertThrows(...);
assertEquals("password must be longer than 6 characters", thrownException.getMessage());
```
Same for other tests

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,52 @@
...
+        ageValidation(user.getAge());
+        validatePassword(user.getPassword());
+        return storageDao.add(user);
+    }
+
+    private void loginValidation(String login) {
+        if (login == null || login.length() <= MINIMAL_LOGIN_LENGTH) {
+            throw new LoginValidateException("your login must be more than 4 characters");
+        } else if (!Character.isLetter(login.charAt(FIRST_LOGIN_CHARACTER_INDEX))) {
+            throw new LoginValidateException("your login should start from the letter");
```
karma-o, line 29: this was not part of the task, as a software engineers we should be mindfull of of solving and implementing business problems, but our own additions should be talked through first, before addition

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,52 @@
...
+        loginValidation(login);
+        if (storageDao.get(login) != null) {
+            throw new LoginValidateException("consumer with " + login + "login is already "
+                    + "exist please try another one login try to login with password");
+        }
+    }
+
+    private void validatePassword(String password) {
+        if (password == null || password.length() <= MINIMAL_PASSWORD_LENGTH) {
+            throw new PasswordValidateException("password must be longer than six characters");
```
karma-o, line 43: replace `six` with a constant

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,52 @@
...
+        } else if (!Character.isLetter(login.charAt(FIRST_LOGIN_CHARACTER_INDEX))) {
+            throw new LoginValidateException("your login should start from the letter");
+        }
+    }
+
+    private void checkExistUserInDataBase(String login) {
+        loginValidation(login);
+        if (storageDao.get(login) != null) {
+            throw new LoginValidateException("consumer with " + login + "login is already "
+                    + "exist please try another one login try to login with password");
```
karma-o, line 37:
```suggestion
            throw new LoginValidateException("consumer with " + login + "login is already existing");
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,52 @@
...
+        }
+    }
+
+    private void validatePassword(String password) {
+        if (password == null || password.length() <= MINIMAL_PASSWORD_LENGTH) {
```
karma-o, line 42: separate null check from the length check to have a seperate exception messages, because right now in case password is null we say that it should be longer than 6 chars, which is misguiding, same for other checks

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,52 @@
...
+            throw new PasswordValidateException("password must be longer than six characters");
+        }
+    }
+
+    private void ageValidation(int age) {
+        if (age < MINIMAL_AGE) {
```
karma-o, line 48: add null check

src/main/java/core/basesyntax/exception/LoginValidateException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.exception;
+
+public class LoginValidateException extends RuntimeException {
```
karma-o, line 3: same as with previus comment about the class name of the exception, class names should be nouns like here:
```suggestion
public class InvalidLoginException extends RuntimeException {
```

src/main/java/core/basesyntax/exception/PasswordValidateException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.exception;
+
+public class PasswordValidateException extends RuntimeException {
```
karma-o, line 3: same
```suggestion
public class InvalidatePasswordException extends RuntimeException {
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+import core.basesyntax.db.Storage;
+import core.basesyntax.exception.InvalidAgeException;
+import core.basesyntax.exception.LoginValidateException;
+import core.basesyntax.exception.PasswordValidateException;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class RegistrationServiceImplTest {
```
karma-o, line 14: still missing test cases that we should add:
* happy case - when user is valid, no validation throws exception, and it gets put in the storage
* null login/age/password - asserts the corresponding exception is thrown
* when trying to register a user, and we already have a user with the same login in the storage - also asserts that exception is thrown. P.S. Definitely check the common mistakes for details on how to put the first user in the `Storage` properly!

karma-o (changes requested):

checklist.md
```diff
@@ -4,13 +4,18 @@
 * Figure out which way is better to initialize your RegistrationServiceImpl class instance: `@BeforeEach` or `@BeforeAll`, what's the difference?
 * Make sure you name your methods according to [convention](https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names).
 * Add tests for all possible User's parameters (null login/password/age, user under 18/18/over 18 years old, negative age, and so on...).
-* Make sure to **not** add something that is not described in the task requirements, or tied to something that may change over time like max user age. It is likely that average and maximum human lifespan will grow over time, so let's keep it in mind.
+* Make sure to **not** add something that is not described in the task requirements, or tied to something that
```
andrii-hoienko, line 7: Please don`t change this file

src/main/java/core/basesyntax/exception/LoginValidateException.java
andrii-hoienko: not resolved

src/main/java/core/basesyntax/exception/PasswordValidateException.java
andrii-hoienko: not resolved

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
andrii-hoienko: Not resolved

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,103 @@
...
+    private User user;
+
+    @BeforeEach
+    void setUp() {
+        user = new User(VALID_LOGIN, VALID_PASSWORD, MINIMAL_AGE);
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.people.clear();
```
andrii-hoienko, line 36: Add this to beforeEach to make sure that db is clean before test execution

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,103 @@
...
+        assertTrue(Storage.people.contains(user));
+    }
+
+    @Test
+    void register_existingLogin_throwsInvalidLoginException() {
+        registrationService.register(user);
+        User newUserWithExistingLogin = new User(VALID_LOGIN, VALID_PASSWORD, MINIMAL_AGE);
+        InvalidLoginException actualException = assertThrows(InvalidLoginException.class,
+                () -> registrationService.register(newUserWithExistingLogin),
+                "consumer with " + user.getLogin() + "login is already existing");
```
andrii-hoienko, line 53: No need to check error message twice

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @AfterEach
+    void tearDown() {
+        Storage.people.clear();
+    }
+
+    @Test
+    void register_validUser_successfullyRegistered() {
+        User registeredUser = registrationService.register(user);
+        assertNotNull(registeredUser);
```
andrii-hoienko, line 42:
```suggestion
```

andrii-hoienko (changes requested):
Some comments are not resolved

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
karma-o: I have seen your message about the primitive type, but this the method that you have written, and if you will go to the `User` class, you will see that it can be null, so let's add a test for it to see the exception that we will get by setting age to null, then change the type of the parameter of this method to wrapper, and add a null check. 
Next time feel free to ask me about the comments that I have left 😉

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,104 @@
...
+    void register_validUser_successfullyRegistered() {
+        User registeredUser = registrationService.register(user);
+        assertEquals(user, registeredUser);
+        assertTrue(Storage.people.contains(user));
+    }
+
+    @Test
+    void register_nullUser_throwsIllegalArgumentException() {
+        user = null;
+        InvalidUserException actualException = assertThrows(InvalidUserException.class,
```
karma-o, line 48: exception in the name of the test and the actual that is thrown are differing

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,104 @@
...
+                () -> registrationService.register(user));
+        assertEquals("User cannot be null",actualException.getMessage());
+    }
+
+    @Test
+    void register_existingLogin_throwsInvalidLoginException() {
+        registrationService.register(user);
```
karma-o, line 55: We should not have used the register method to put the first user in the storage, because this method is under a test at this actual moment and can not be used to arrange the test data, because it is not guaranteed to work properly as expected, better to use low-level approaches like putting him there directly via `storageDao` or `Storage`
This is the point from common mistakes, check it for more details

checklist.md
```diff
@@ -4,13 +4,18 @@
...
 In this case would be good to test: empty password (0-length - not-ok), 'abc' (3-length - not-ok), 'abcdf' (5-length - **not-ok edge case**), 
 'abcdef' (6-length - **ok edge case**), 'abcdefgh' (8-length - ok). Few not-ok checks you can have in one not-ok test method. Same for ok-test.
 Testing only 'abc' (3-length - not-ok) and 'abcdefgh' (8-length - ok) is not enough because we do not check how the program behave with edge cases.
 * Make sure that after successful registration user was added to storage and `register` method returned correct user.
-* Use low-level methods like `Storage.people.add()` directly instead of `RegistrationService.register()` method to add users for testing. Because we can not use the same method which is under a test to insert some mock test data into our storage since we can not guarantee its correctness yet.
+* 
+* Use low-level methods like `Storage.people.add()` directly instead of `RegistrationService.register()`
+method to add users for testing. Because we can not use the same method which is under a test to insert 
+some mock test data into our storage since we can not guarantee its correctness yet.
+*
```
karma-o, line 18: here it is, you even changed it, BTW we can not do that and should revert changes to this file

karma-o (approved):
Approved but left a very important comment, please check it

PR #936, opened 2024-03-10
Created registration validation.

Recdt:


Eidenn005 (approved):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,67 @@
...
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exceptions.RegistrationFailureException;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
-    private final StorageDao storageDao = new StorageDaoImpl();
+    private static final int MIN_DATA_LENGTH = 6;
+    private static final int MIN_AGE = 18;
+    private StorageDao storageDao = new StorageDaoImpl();
```
karma-o, line 11: why storage is n ot not final anymore?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,67 @@
...
 
     @Override
     public User register(User user) {
-        return null;
+        userNullDataValidation(user);
+        userLoginAndPasswordValidation(user);
+        ageValidation(user);
```
karma-o, line 17: The order of those validations is a bit messy, methods should have as isolated and single responsibility as possible, so usually we do group several different fields by a single validation criteria, so let's add couple methods like `validateLogin()` `validatePassword()` `validateAge()` and we can call them in some `validateUser()` instead of what we have here

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,67 @@
...
+
+        String login = user.getLogin();
+        if (storageDao.get(login) != null) {
+            throw new RegistrationFailureException("This user login exists: " + login);
+        }
+
+        storageDao.add(user);
+        return storageDao.get(login);
```
karma-o, line 25: I think `add()` also returns that user, so there is no need to call `get()` after it

src/test/java/core/basesyntax/HelloWorldTest.java
```diff
@@ -1,8 +1,145 @@
...
+        User user = null;
+        assertThrows(RegistrationFailureException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void register_addIdenticalUsers_notOk() {
+        User user = new User();
+        user.setLogin("Antonio");
+        user.setPassword("12345678");
+        user.setAge(23);
```
karma-o, line 65: couple of moments here:
* there is not need to create a new user and populate it with data by hand each time, add a `validUser` field  in the test class, and initialize it with the valid fields `BeforeEach` test. Then when you need to assert a validation error, just set a single not valid field to it
* magic values should be constants, same for each other test method

src/test/java/core/basesyntax/HelloWorldTest.java
```diff
@@ -1,8 +1,145 @@
...
+
+        registrationService.register(user);
+        assertThrows(RegistrationFailureException.class, () -> registrationService.register(user));
```
karma-o, line 68: we can also assert the message that we have in that exception, so let's add it to all similar tests

src/test/java/core/basesyntax/HelloWorldTest.java
```diff
@@ -1,8 +1,145 @@
...
+    }
+
+    @Test
+    void register_addIdenticalUsers_notOk() {
+        User user = new User();
+        user.setLogin("Antonio");
+        user.setPassword("12345678");
+        user.setAge(23);
+
+        registrationService.register(user);
```
karma-o, line 67: this part of inserting the first user, is an arrangement of the test data, which should not be done at least with the same method that is under the test, because we are not sure that it is working as intended at the moment. So we should use some other, more low-level approaches here.
check common mistakes

src/test/java/core/basesyntax/HelloWorldTest.java
```diff
@@ -1,8 +1,145 @@
...
+
+        assertThrows(RegistrationFailureException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void register_allParametersNull_notOk() {
+        User user = new User();
+        user.setLogin(null);
+        user.setPassword(null);
+        user.setAge(null);
```
karma-o, line 141: wouldn't say that we need that case because null login and lull password are already tested separately, we are only missing a separate null check of age test.
What else is interesting here:
* The exception about null age will never be thrown as you are expecting it, and password in fact too, because it will always check the login first, and thus will always throw the same exception about invalid login

We should strive for as deep & detailed tests as possible, to guarantee that the code we wrote works **exactly** as we think it does

karma-o (changes requested):

src/test/java/core/basesyntax/HelloWorldTest.java
Recdt: I think in java we can't add message to assertThrows?

src/test/java/core/basesyntax/HelloWorldTest.java
```diff
@@ -1,8 +1,128 @@
...
+    void register_passwordEdges_Ok() {
+        validUser.setLogin("Marco1");
+        validUser.setPassword("123456");
+        User result = registrationService.register(validUser);
+        assertEquals(result, validUser, "User and added result aren't the same.");
+
+        validUser = new User();
+        validUser.setLogin("Parmezan");
+        validUser.setPassword("12345678910");
+        validUser.setAge(VALID_AGE);
```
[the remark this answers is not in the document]
Recdt: I can't do that. Because if it's ok test, I am adding user to DataBase, and I can't add the same user to it. And if I will change some fields without creating new user I will change the user in DataBase by reference

src/test/java/core/basesyntax/HelloWorldTest.java
```diff
@@ -1,8 +1,128 @@
...
+        validUser = null;
+        assertThrows(RegistrationFailureException.class,
+                () -> registrationService.register(validUser),
+                "Invalid result for null user.");
+    }
+
+    @Test
+    void register_addIdenticalUsers_notOk() {
+        validUser.setLogin("Antonio");
+        Storage.people.add(validUser);
```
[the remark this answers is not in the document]
Recdt: ?

src/test/java/core/basesyntax/HelloWorldTest.java
```diff
@@ -1,8 +1,134 @@
...
+import core.basesyntax.service.RegistrationServiceImpl;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
 /**
  * Feel free to remove this class and create your own.
  */
 public class HelloWorldTest {
```
karma-o, line 19:
```suggestion
public class RegistrationServiceTest {
```

karma-o (approved):
Nice! :fire:

PR #939, opened 2024-03-13
changed User.java(add constructor) and made the task

esidarapp:


Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -9,6 +10,35 @@ public class RegistrationServiceImpl implements RegistrationService {
...
+
+    public boolean validate(User user) throws InvalidDataException {
+        if (user.getAge() < 18) {
+            throw new InvalidDataException("User`s age must be at least 18!");
+        }
+        if (user.getLogin().length() < 6) {
+            throw new InvalidDataException("User`s login length must be at least 6 symbols!");
+        }
+        if (user.getPassword().length() < 6) {
+            throw new InvalidDataException("User`s password length must be at least 6 symbols!");
```
okuzan, line 36: don't hardcode in these messages, use constants

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -9,6 +10,35 @@ public class RegistrationServiceImpl implements RegistrationService {
...
+        try {
+            if (validate(user)) {
+                storageDao.add(user);
+                return user;
+            } else {
+                return null;
+            }
+        } catch (InvalidDataException e) {
+            throw new RuntimeException(e);
+        }
```
okuzan, line 25: returning null is a bad practice in most cases
validate could as well be void, it's a guardian method, if smth is wrong - it will throw exception, if not - proceed with saving user

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,90 @@
...
+        assertThrows(InvalidDataException.class, () -> registrationService.validate(user));
+        user.setLogin("12345");
+        assertThrows(InvalidDataException.class, () -> registrationService.validate(user));
+    }
+
+    @Test
+    public void validate_PasswordMoreThan6_Ok() {
+        User userOK = new User("11111111", "222222", 25);
+        assertDoesNotThrow(() -> registrationService.validate(userOK));
+    }
```
okuzan, line 62: you need only one _ok test, where valid data passes

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Test
+    public void validate_LoginMoreThan6_Ok() {
+        User userOK = new User("11111111", "222222", 25);
+        assertDoesNotThrow(() -> registrationService.validate(userOK));
+    }
+
+    @Test
+    public void validate_LoginLessThan6_NotOk() {
```
okuzan, line 49:
```suggestion
    public void validate_LoginTooShort_NotOk() {
```
you may forget to change method name if constraint changes, better to be less specific

okuzan (changes requested):
You should test age 18 - edge case

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -6,6 +6,14 @@
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_LENGTH = 6;
+    private static final int MIN_AGE = 18;
+    private static final String AGE_MESSAGE = "User's age must be at least 18!";
+    private static final String LOGIN_LENGTH_MESSAGE = "User's login length"
+            + " must be at least 6 symbols!";
+    private static final String PASSWORD_LENGTH_MESSAGE = "User's password length "
+            + "must be at least 6 symbols!";
```
okuzan, line 15: this doesn't fix the issue - you hardcode 6. Use constant MIN_LENGTH

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,47 @@
...
+        if (user == null) {
+            throw new InvalidDataException("User cannot be null");
+        }
+        try {
+            validate(user);
+            storageDao.add(user);
+            return user;
+        } catch (InvalidDataException e) {
+            throw new RuntimeException(e);
+        }
```
okuzan, line 30: remove try catch
your method shouldn't throw Runtime, but your custom exception

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    public void register_NullUser_NotOk() {
+        assertThrows(InvalidDataException.class, () -> registrationService.register(null));
+    }
+
+    @Test
+    public void register_NotNullUser_Ok() {
+        User registeredUser = registrationService.register(new User("logggg", "passsssw", 22));
+        assertNotNull(registeredUser);
+    }
```
okuzan, line 77: only one _ok test is needed
test age of 18

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+        user.setAge(0);
+        assertThrows(InvalidDataException.class, () -> registrationService.validate(user));
+        user.setAge(-100);
+        assertThrows(InvalidDataException.class, () -> registrationService.validate(user));
+        user.setAge(17);
+        assertThrows(InvalidDataException.class, () -> registrationService.validate(user));
+    }
+
+    @Test
+    public void validate_LoginLength_NotOk() {
```
okuzan, line 43: loginLength what? tooShort? valid?

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,47 @@
...
+    }
+
+    public void validate(User user) throws InvalidDataException {
```
okuzan, line 33: we don't declare throwing unchecked exceptions
why is it public? shouldn't be exposed, it's utility for main public method register

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+        assertThrows(InvalidDataException.class, () -> registrationService.validate(user));
+        user.setPassword("12345");
+        assertThrows(InvalidDataException.class, () -> registrationService.validate(user));
+    }
+
+    @Test
+    public void validate_UserWithSuchLoginAlreadyExists_NotOk() {
+        User newUser = new User("testuser","newpassword", 19);
+        assertThrows(InvalidDataException.class, () -> registrationService.validate(newUser));
+    }
```
okuzan, line 66: we don't test private methods (which validate should be), only public methods

okuzan (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,43 @@
...
+        }
+    }
+
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new InvalidDataException("User cannot be null");
+        }
+        validate(user);
```
okuzan, line 39: this `if` should be part of `validate()`

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+    private RegistrationServiceImpl registrationService;
+    private StorageDaoImpl storageDao;
+    private User user;
+
+    @BeforeEach
+    public void setUp() {
+        registrationService = new RegistrationServiceImpl();
+        storageDao = new StorageDaoImpl();
+        user = new User("testuser", "password123", 20);
+        storageDao.add(user);
```
okuzan, line 22: we only need this for test with duplicate login, so move it there

okuzan (approved):
Good job

PR #940, opened 2024-03-13
implemented and tested RegistrationService

ghost:


src/main/java/core/basesyntax/model/User.java
```diff
@@ -8,6 +8,13 @@ public class User {
     private String password;
     private Integer age;
 
+    public User(Long id, String login, String password, Integer age) {
+        this.id = id;
```
Vasul-Pankiv, line 12: Do not set Id through the constructor. There are cases when user is not saved in db and does not have any id

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,147 @@
...
+        service = new RegistrationServiceImpl(storageDao);
+    }
+
+    @AfterEach
+    void tearDown() {
+        Storage.people.clear();
+    }
+
+    @Test
+    void register_notNull_ok() {
```
Vasul-Pankiv, line 37: also in ok cases check storage, size should increase and assertEquals for register() return value

Vasul-Pankiv (changes requested):
Good job

kozhukhovsky (approved):

PR #951, opened 2024-04-07
First attempt

jszymanski97:
Not too keen on using different adnotations ie. @BeforeEach @AfterEach etc. as it was not explaind at all in course videos and the theory explains very little about it...  So I used @Test everywhere.

### DIFF
--- a/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
+++ b/src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
@@ -2,13 +2,57 @@ package core.basesyntax.service;
 
 import core.basesyntax.dao.StorageDao;
 import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.db.Storage;
 import core.basesyntax.model.User;
 
 public class RegistrationServiceImpl implements RegistrationService {
+
+    private static final int LOGIN_MIN_LENGTH = 6;
+    private static final int PASSWORD_MIN_LENGTH = 6;
+    private static final int USER_MIN_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
-    public User register(User user) {
-        return null;
+    public User register(User user) throws ValidationException {
+        if (user == null) {
+            throw new ValidationException("Can't pass Null");
+        }
+
+        if (user.getId() == null) {
+            throw new ValidationException("ID is required");
+        }
+
+        if (user.getLogin() == null) {
+            throw new ValidationException("Login is required");
+        }
+
+        if (user.getPassword() == null) {
+            throw new ValidationException("Password is required");
+        }
+
+        if (user.getAge() == null) {
+            throw new ValidationException("Age is required");
+        }
+
+        if (Storage.people.contains(user)) {
+            throw new ValidationException("This user already exists");
+        }
+
+        if (user.getLogin().length() < LOGIN_MIN_LENGTH) {
+            throw new ValidationException("Login must be at least " + LOGIN_MIN_LENGTH
+                    + " characters long");
+        }
+
+        if (user.getPassword().length() < PASSWORD_MIN_LENGTH) {
+            throw new ValidationException("Password must be at least " + PASSWORD_MIN_LENGTH
+                    + " characters long");
+        }
+
+        if (user.getAge() < USER_MIN_AGE) {
+            throw new ValidationException("User must be at least " + USER_MIN_AGE + " years old");
+        }
+
+        storageDao.add(user);
+        return user;
     }
 }
--- a/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
+++ b/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
@@ -1,0 +1,123 @@
+package core.basesyntax.service;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,57 @@
...
+
+        if (user.getPassword() == null) {
+            throw new ValidationException("Password is required");
+        }
+
+        if (user.getAge() == null) {
+            throw new ValidationException("Age is required");
+        }
+
+        if (Storage.people.contains(user)) {
```
d1sam, line 37: you should use `StorageDao`

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,57 @@
...
+
+    private static final int LOGIN_MIN_LENGTH = 6;
+    private static final int PASSWORD_MIN_LENGTH = 6;
+    private static final int USER_MIN_AGE = 18;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
-    public User register(User user) {
-        return null;
+    public User register(User user) throws ValidationException {
```
d1sam, line 16: can we register user with spaces inside username?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
d1sam, line 1: you should check equality of users' objects, not User.class.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,123 @@
...
+    @Test
+    void loginLength_3Length_notOk() {
+        User user = new User(123L, "sho", "correctPassword", 20);
+        assertThrows(ValidationException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    void loginLength_5Length_notOk() {
+        User user = new User(123L, "short", "correctPassword", 20);
+        assertThrows(ValidationException.class, () -> registrationService.register(user));
```
d1sam, line 43:
```suggestion
        assertThrows(ValidationException.class, () -> registrationService.register(user), "If user login is short it should throw ValidationException Exception!");
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
d1sam: I saw your comment about `@BeforeEach, @AfterEach, @BeforeAll, @AfterAll.` I still advise you to deal with them, because they are actually very important annotations and will significantly reduce the test code and make it more readable.

d1sam, line 1: please avoid String literals

d1sam (changes requested):
Good job, but there are some changes, that should be done according my comments.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
jszymanski97: Should I change all messages thrown in my ValidExceptions to constains ? so Instead having
if (user.getId() == null) {
            throw new ValidationException("ID is required");
        }

I should create a:

private static final String ID_REQUIRED = "ID is required";
if (user.getId() == null) {
            throw new ValidationException(ID_REQUIRED);
        }

Or do it this way in future codes... ?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
nowickimj: @jszymanski97 a message for the violation of assertion

nowickimj: @jszymanski97 Personally I think such message is redundant here - the reason for throwing an exception is provided in a test method's name.

IMO your tests could be improved by evaluating whether the message of thrown exception matches the expected value from `ErrorMessage` class for each case.

nowickimj (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
jszymanski97: Okay, Thank you I think I get the point now… Checking if correct message is returned instead correct exception. Will get these corrected later on today
Thank You

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,155 @@
...
+                () -> registrationService.register(user));
+        assertEquals(ErrorMessage.ERROR_UNDERAGE_USER, actual.getMessage());
+    }
+
+    @Test
+    void spaceInLogin_spaceLogin_notOk() {
+        User user = new User(123L, "space Login", "correctPassword", 20);
+        ValidationException actual = assertThrows(ValidationException.class,
+                () -> registrationService.register(user));
+        assertEquals(ErrorMessage.ERROR_SPACE_IN_LOGIN, actual.getMessage());
```
nowickimj, line 153: Lots of unit tests in this class use the same logic - how about extracting it to a private method to simplify the code? Something like:
```java
private void assertUserRegistrationThrowsValidationExceptionWithMessage(User user, String errorMessage);
```

nowickimj (changes requested):
Small refactoring to remove code duplicates

nowickimj (approved):

PR #952, opened 2024-04-14
done

SukhachovS:


src/test/java/core/basesyntax/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,144 @@
...
+        assertDoesNotThrow(() -> registrationService.register(user));
+    }
+
+    @Test
+    public void register_minimumValidAge_Ok() {
+        user.setLogin("testUserMinAge");
+        user.setPassword("testPassword");
+        user.setAge(18);
+        assertDoesNotThrow(() -> registrationService.register(user),
+                "Registration using minimum valid age should be successful");
```
olekskov, line 37: It's reraly required to use assertDoesNotThrow
If logic in test throws exception that is not expected with assertThrows then test is automatically fail. What is same as using assertDoesNotThrow
```suggestion
        registrationService.register(user);
```

olekskov (approved):
Good job

PR #959, opened 2024-04-27
registration validation

BelitskyiAlexandr:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,32 @@
...
+            throw new InvalidDataException("User already exists");
+        }
+        if (user.getLogin().length() < LOGIN_PASS_LENGTH) {
+            throw new InvalidDataException("Login must be at least 6 characters length");
+        }
+        if (user.getPassword().length() < LOGIN_PASS_LENGTH) {
+            throw new InvalidDataException("Password must be at least 6 characters length");
+        }
+        if (user.getAge() < ACCEPTABLE_AGE) {
+            throw new InvalidDataException("Age must be at least 18");
```
okuzan, line 28: don't hardcode, you have constants for this. This code is fragile - for example, if you later on change min length to 8, you may forget to change exception messages, and it will confuse users

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+    @BeforeEach
+    void setUp() {
+        user = new User();
+        user.setLogin("loogiin");
+        user.setPassword("password");
+        user.setAge(20);
+    }
+
+    @Test
+    void register_Ok() {
```
okuzan, line 31:
```suggestion
    void register_validUser_ok() {
```
Please make all method names adhere to the naming conventions outlined in task description

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+    void register_Exists() {
+        User sameUser = user;
+        assertThrows(InvalidDataException.class, () -> {
+            registrationService.register(sameUser);
+        });
+    }
+
+    @AfterAll
+    static void afterAll() {
+        registrationService = new RegistrationServiceImpl();
```
okuzan, line 64: not sure why is this in afterAll
better clean up the storage to not share state between tests
image

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+    void setUp() {
+        user = new User();
+        user.setLogin("loogiin");
+        user.setPassword("password");
+        user.setAge(20);
+    }
+
+    @Test
+    void register_Ok() {
+        assertTrue(registrationService.register(user) == user);
```
okuzan, line 32: we have assertEquals for this

okuzan: you also need to make sure not only that return value is correct, but that is was added to the storage

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+        });
+    }
+
+    @Test
+    void register_Exists() {
+        User sameUser = user;
+        assertThrows(InvalidDataException.class, () -> {
+            registrationService.register(sameUser);
+        });
+    }
```
okuzan, line 60: why does it even throw exception? Your tests aren't independent, if you only run this test, it will fail

okuzan (changes requested):
Make sure to read the checklist of common mistakes first. Test edge cases - length of 5/6, age of 17/18, fix naming

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
BelitskyiAlexandr: Underscores may appear in JUnit test method names to separate logical components of the name, with each component written in [lowerCamelCase](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case), for example transferMoney_deductsFromSource. _There is no One Correct Way to name test methods._ - so I named it how I see

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
BelitskyiAlexandr: How can I do this?
Write a method of getting the length of the db in the registration service - it is not logical.
It is also impossible to access the db field from the outside

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+    @AfterEach
+    void tearDown() {
+        Storage.people.clear();
+    }
+
+    @AfterAll
+    static void afterAll() {
```
Elena-Bruyako, line 153: Redundant

Elena-Bruyako (changes requested):
Rename all test methods according to the common mistakes file. 
Don't use numbers in naming, it's better MinAge or MinPasswordLength

BelitskyiAlexandr:
> Rename all test methods according to the common mistakes file. Don't use numbers in naming, it's better MinAge or MinPasswordLength

image

https://github.com/mate-academy/jv-registration-validation/pull/959#discussion_r1581895253

Please, give me more accurate instructions

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @Test
+    void register_validUser_ok() {
+        assertEquals(registrationService.register(user), user);
+        assertEquals(Storage.people.size(), 1);
+    }
+
+    @Test
+    void register_login_notOk() {
```
Elena-Bruyako, line 38: Best practice it to isolate test methods, create one method for particular test case. 
If you want to test several conditions or options for one method, you can use `@ParameterizedTest`.
When your test 'register_login_notOk' fails, you won't know which condition is not working.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @BeforeAll
+    static void beforeAll() {
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @BeforeEach
+    void setUp() {
+        user = new User();
+        user.setLogin("loogiin");
```
Elena-Bruyako, line 26: Best practice is to set up data in each test case: given (user), when, then.

Elena-Bruyako (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
BelitskyiAlexandr: Why don't we know? There we can see which assert fails and row number:
image

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
BelitskyiAlexandr: But we need some kind of basic user for most tests - and this is +3-4 lines of code in the method. In my tests, this is more than 30 lines of repeated code.

Can you please provide a link to the best practices?

sarakhmen: @BelitskyiAlexandr if you found that this way you can avoid a lot of code duplication and you're sure that this test user wouldn't change then it's ok to have it initialized in beforeeach as you've done

sarakhmen (approved):
LGTM!

PR #965, opened 2024-05-10
Added TestClass. Implements method registration. Created new tests fo…

VadymOziPantielieienko:
…r all necessary parameters.

### DIFF
--- a/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
+++ b/src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
@@ -1,0 +1,214 @@
+package core.basesyntax.service;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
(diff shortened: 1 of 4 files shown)


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,214 @@
...
+        setNotValidData();
+    }
+
+    @BeforeEach
+    void beforeEach() {
+        testUser = new User();
+    }
+
+    @Test
+    void registerOneUser_OK() {
```
Elena-Bruyako, line 41: it's better to stick to the such test rule for each test case:
given - create test data 
when - call method that you would like to test
then - compare results

It makes tests more readable and avoid some unnecessary logic (or for loop like in your case).

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #968, opened 2024-05-17
solved task

Yaromyros:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,23 @@
...
 
 public class RegistrationServiceImpl implements RegistrationService {
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new InvalidDataException("There already is user with this login!");
+        } else if (user.getLogin().length() < 6) {
```
Elena-Bruyako, line 15: Make 6 and 18 constants. 
It's better to add checking: password and login not null

src/test/java/core/basesyntax/RegistrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+import core.basesyntax.db.Storage;
+import core.basesyntax.exception.InvalidDataException;
+import core.basesyntax.model.User;
+import core.basesyntax.service.RegistrationService;
+import core.basesyntax.service.RegistrationServiceImpl;
+import java.util.Random;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class RegistrationTest {
```
Elena-Bruyako, line 19: It's better to rename and implement test cases only for methods in the RegistrationService class. 
```suggestion
class RegistrationServiceTest {
```

src/test/java/core/basesyntax/RegistrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+        assertEquals(userTest, retrievedUser, "Storage does not contain user");
+    }
+
+    @Test
+    public void test_storageUser_returnNull_isOk() {
+        User retrievedUser = storageDao.get("Coco123");
+        assertNull(retrievedUser, "Storage should return null to un-existing user");
+    }
+
+    /* Tests in class User */
```
Elena-Bruyako, line 67: We don't need to write tests to check that Java works well.

src/test/java/core/basesyntax/RegistrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+        assertEquals("newUser", user.getLogin());
+        assertEquals("newPassword", user.getPassword());
+        assertEquals(25, user.getAge());
+        assertEquals(2, user.getId());
+    }
+
+    /* Tests in class RegistrationServiceImpl */
+
+    @Test
+    public void test_checking_user_OnExisting() {
```
Elena-Bruyako, line 156: Let's stick to the rules: 
There are a lot of ways to name your test methods. The main point is that they should have informative names and be consistent along with other developers in your team. For this task use such convention: <methodUnderTest>_<state>_<expectedBehavior>; For example, if we are testing the method register with a null
user's age the test method name should be register_nullAge_notOk. notOk is because the test expects the register method to throw an exception.

src/test/java/core/basesyntax/RegistrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+    @Test
+    public void test_age_less18_notOk() {
+        User user = new User();
+        user.setLogin("PON");
+        user.setPassword("pon");
+        user.setAge(9);
+
+        assertThrows(InvalidDataException.class, () -> registrationService.register(user));
+    }
```
Elena-Bruyako, line 199: Please, add test cases according to the checklist and comment in RegistrationServiceImpl 
Test cases: 
password / login == null
valid user -> ok

Read [checklist](https://github.com/mate-academy/jv-registration-validation/blob/master/checklist.md) file for more information

Elena-Bruyako (changes requested):

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,320 @@
...
+    private static RegistrationService registrationService;
+
+    @BeforeAll
+    public static void setUp() {
+        random = new Random();
+        userTest = new User();
+        userTest.setAge(MIN_AGE);
+        userTest.setId(30L);
+        userTest.setLogin("CoolGuy");
+        userTest.setPassword("Qwerty123_4");
```
okuzan, line 38: no need to set Id, it will be set by the database
if you always use those same setters together, consider creating a constructor

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,320 @@
...
+            User eachAddedUser = storageDao.add(user);
+
+            assertEquals(eachAddedUser, user, "Not all users were added to the storage");
+        }
+    }
+
+    @Test
+    public void test_storageUser_containsUser() {
+        User retrievedUser = storageDao.get("CoolGuy");
+        assertEquals(userTest, retrievedUser, "Storage does not contain user");
```
okuzan, line 65: tests should be independent - do you rely on the other tests to fill the data?

src/test/java/core/basesyntax/RegistrationServiceTest.java
okuzan, line 38: I would recommend recreating this test user in beforeEach, not beforeall. to ensure it hasn't been corrupted

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,320 @@
...
+        user1.setLogin("user1");
+        user1.setPassword("password");
+        user1.setAge(MIN_AGE);
+
+        User user2 = new User();
+        user2.setLogin("user1");
+        user2.setPassword("password");
+        user2.setAge(MIN_AGE);
+
+        assertEquals(user1, user2);
```
okuzan, line 82: it doesn't test functionality of registration service, remove this test

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,320 @@
...
+        user1.setLogin("user1");
+        user1.setPassword("password");
+        user1.setAge(MIN_AGE);
+
+        User user2 = new User();
+        user2.setLogin("user2");
+        user2.setPassword("password");
+        user2.setAge(MIN_AGE);
+
+        assertNotEquals(user1, user2);
```
okuzan, line 97: same
we don't test model classes like user in service tests

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,320 @@
...
+        user1.setLogin("user1");
+        user1.setPassword("password");
+        user1.setAge(MIN_AGE);
+
+        User user2 = new User();
+        user2.setLogin("user1");
+        user2.setPassword("password");
+        user2.setAge(MIN_AGE);
+
+        assertEquals(user1.hashCode(), user2.hashCode());
```
okuzan, line 112: same

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,320 @@
...
+        user1.setLogin("user1");
+        user1.setPassword("password");
+        user1.setAge(MIN_AGE);
+
+        User user2 = new User();
+        user2.setLogin("user2");
+        user2.setPassword("password");
+        user2.setAge(MIN_AGE);
+
+        assertNotEquals(user1.hashCode(), user2.hashCode());
```
okuzan, line 127: same

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    @Test
+    public void test_gettersAndSetters_workOk() {
```
okuzan, line 131: same

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,320 @@
...
+        user1.setPassword("password");
+        user1.setAge(MIN_AGE);
+        registrationService.register(user1);
+
+        User user2 = new User();
+        user2.setLogin("Recruit");
+        user2.setPassword("KingBoss");
+        user2.setAge(MIN_AGE);
+
+        assertThrows(InvalidDataException.class, () -> registrationService.register(user2));
```
okuzan, line 169: checklist issue, you can't rely on the method you test in the test of this method 🤯
image

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    @Test
+    public void register_loginLength_lessThan6_throwsException() {
+        User user = new User();
+        user.setLogin("PON");
```
okuzan, line 175: test length of exactly 5, pre-edge case

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,320 @@
...
+        user.setPassword("MyCode");
+        user.setAge(MIN_AGE);
+
+        assertThrows(InvalidDataException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    public void register_loginLength_6_isOK() {
+        User user = new User();
```
okuzan, line 184: you need only one _ok test. When every field is valid - it works. That's it

src/test/java/core/basesyntax/RegistrationServiceTest.java
```diff
@@ -0,0 +1,320 @@
...
+        user.setPassword("passCodes");
+        user.setAge(MIN_AGE);
+
+        assertDoesNotThrow(() -> registrationService.register(user));
+    }
+
+    @Test
+    public void register_loginLength_moreThan6_isOK() {
+        User user = new User();
+        user.setLogin("NancyHale");
```
okuzan, line 195: test length of exactly 6
same for passwords - test 5/6, as you do for age - 17/18

okuzan (changes requested):

src/test/java/core/basesyntax/model/UserTest.java
```diff
@@ -0,0 +1,66 @@
+package core.basesyntax.model;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+import org.junit.jupiter.api.Test;
+
+class UserTest {
```
Elena-Bruyako, line 8: It's redundant test class. As I see you are trying to test if Java methods work well, there is no logic to test.

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,189 @@
...
+        User user = new User("derek", "derekD12", MIN_AGE);
+
+        assertThrows(InvalidDataException.class, () -> registrationService.register(user));
+    }
+
+    @Test
+    public void register_loginLength_6_registrationSuccessful() {
+        User user = new User("Peters", "passCodes", MIN_AGE);
+
+        assertDoesNotThrow(() -> registrationService.register(user));
```
Elena-Bruyako, line 58:
```suggestion
        User actual = registrationService.register(user);
        assertEquals(user, actual);
```

The same for other OK cases. BTW you can write only one valid test case

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,33 @@
...
+            throw new InvalidDataException("There already is user with this login!");
+        } else if (user.getLogin().length() < ALLOWED_LENGTH) {
+            throw new InvalidDataException("Login is less than 6 characters, try longer!");
+        } else if (user.getPassword().length() < ALLOWED_LENGTH) {
+            throw new InvalidDataException("Password is less than 6 characters, try longer!");
+        } else if (user.getAge() < MIN_AGE && user.getAge() >= 0) {
+            throw new InvalidDataException("User is under 18!");
+        } else if (user.getAge() < 0) {
+            throw new InvalidDataException("Age can not be negative!");
+        }
```
okuzan, line 31: remember, that we don't need else after returning/throwing

okuzan (approved):
Good job

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
Yaromyros: Understood, big thanks!

PR #981, opened 2024-06-09
First solution

Wojtas1996PL:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,40 @@
...
 
     @Override
-    public User register(User user) {
-        return null;
+    public User register(User user) throws RegistrationException {
+        for (User userInLoop : Storage.people) {
+            if (userInLoop.equals(user)) {
+                throw new RegistrationException("User already exists");
+            }
+        }
```
MateuszRostek, line 19: Using a `for` loop is redundant in this case, it should be done like this (for example):
```suggestion
        if (storageDao.get(user.getLogin()) != null) {
            throw new RegistrationException("Login: " + user.getLogin()
                    + " is already taken! Choose another one");
        }
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+        User user = new User();
+        user.setLogin(AVERAGE_LOGIN);
+        user.setPassword(AVERAGE_PASSWORD);
+        user.setAge(AGE_OF_TWENTY);
+        registrationService.register(user);
+        for (int i = 0; i < 5; i++) {
+            assertThrows(RegistrationException.class, () -> registrationService.register(user));
+        }
+
+    }
```
MateuszRostek, line 34: This test should change right now due to changes in the service. Also you are missing the `_notOk` at the end.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
MateuszRostek, line 1: All test methods should start with `register` not with `user`:
image

MateuszRostek: Add more test for edge cases and for successful ones:

image

image

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Test
+    void user_nullAge_notOk() {
+        User user = new User();
+        user.setAge(null);
```
MateuszRostek, line 39: In this case all fields are `null` by default so you tests check what if all fields are null, you should have all fields filled with some dummy data, and then set a field to null, so you test what happens when all fields are correct and only one is null.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+        registrationService = new RegistrationServiceImpl();
+    }
+
+    @Test
+    void user_alreadyExist() throws RegistrationException {
+        User user = new User();
+        user.setLogin(AVERAGE_LOGIN);
+        user.setPassword(AVERAGE_PASSWORD);
+        user.setAge(AGE_OF_TWENTY);
+        registrationService.register(user);
```
MateuszRostek, line 29: image

MateuszRostek (changes requested):
Before creating a PR go through ALL the `COMMON MISTAKES` provided at the end of the task description:
https://github.com/mate-academy/jv-registration-validation/blob/master/checklist.md

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,257 @@
...
+        user3.setPassword(GOOD_PASSWORD_THIRD);
+        user1.setAge(AGE_OF_EIGHTEEN);
+        user2.setAge(AGE_OF_TWENTY);
+        user3.setAge(AGE_OF_FORTY);
+        registrationService.register(user1);
+        registrationService.register(user2);
+        registrationService.register(user3);
+        assertEquals(AGE_OF_EIGHTEEN, user1.getAge());
+        assertEquals(AGE_OF_TWENTY, user2.getAge());
+        assertEquals(AGE_OF_FORTY, user3.getAge());
```
MateuszRostek, line 134: in the `_ok` tests you should compare the users you created manually with what you get from the `registrationService.register(someUser)`. For example:

`assertEquals(user1, registrationService.register(user1))`

PR #982, opened 2024-06-09
jv-registration-validation solution

c-kalenyk:


src/main/java/core/basesyntax/service/RegistrationServiceException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.service;
+
+public class RegistrationServiceException extends RuntimeException {
```
Elena-Bruyako, line 3:
```suggestion
public class RegistrationException extends RuntimeException {
```

Exception name should be more abstract

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,31 @@
...
+        }
+        if (user.getPassword() == null || user.getPassword().length() < MINIMAL_LENGTH) {
+            throw new RegistrationServiceException(
+                    "password must consist of at least 6 characters");
+        }
+        if (user.getAge() == null || user.getAge() < MINIMAL_AGE) {
+            throw new RegistrationServiceException(
+                    "registration is not available for users under 18");
+        }
+        if (doesLoginExists(user.getLogin())) {
```
Elena-Bruyako, line 26:
```suggestion
        if (storageDao.get(login) != null) {
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -5,10 +5,31 @@
...
+            throw new RegistrationServiceException("user with such login already exists");
+        }
+        return storageDao.add(user);
+    }
+
+    private boolean doesLoginExists(String login) {
```
Elena-Bruyako, line 32: Don't need to create separate method here

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,185 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static RegistrationService registrationService;
+    private StorageDao storageDao;
+    private User actualUser;
+    private User actualUser2;
```
Elena-Bruyako, line 19: Numbers in naming is a bad approach

Elena-Bruyako: Do you really need two users for test? Just asking your opinion about this

Elena-Bruyako (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
c-kalenyk: It's just the only way i see to check if method will behave correctly when there will be more than 1 element. For example if it will throw an exception then such login already exists or return correct user from storage. I've tried to just change actualUser and add it again, but seems like it doesn't work this way.

Elena-Bruyako: For checking if user with the same login already exists, you could create User, add it to storage.people, create newUser with the same login and call register method, that throws an exception

Elena-Bruyako (changes requested):
Answered to you comment

sarakhmen (approved):
good job!

PR #1002, opened 2024-07-24
Hw registration validation

swoadie:
Implemented the registration service and wrote unit tests for it.


src/test/java/core/basesyntax/service/RegistrationServiceTest.java
```diff
@@ -0,0 +1,123 @@
...
+        assertEquals("Password cannot be null!", exception.getMessage());
+    }
+
+    @Test
+    void register_lowAge_notOk() {
+        User userLowAge = new User();
+        userLowAge.setLogin(VALID_LOGIN);
+        userLowAge.setPassword(VALID_PASSWORD);
+        userLowAge.setAge(INVALID_BORDERLINE_AGE);
+        userLowAge.setAge(INVALID_NEGATIVE_AGE);
```
MentorMateAcademy, line 105: why do you reset age?

MentorMateAcademy (changes requested):

src/test/java/core/basesyntax/service/RegistrationServiceTest.java
swoadie: @MentorMateAcademy in the file Common mistakes (jv-registration-validation) it was indicated that I can have several not-ok checks in one not-ok test method. Should I remove them?

Elena-Bruyako: Correct, but in this case you need to create several users with different not ok ages 
Now in your case you just set up age and then reset different age

Elena-Bruyako (changes requested):

sarakhmen (approved):

PR #1004, opened 2024-07-31
Task done

nastiapyshak:


okuzan (changes requested):
we need only one _ok test, where everything is valid
you should test age 17 exactly - it's an edge case, also password and login length of exactly 5 and 6

sarakhmen (changes requested):
you didn't test all edge cases as advised by the previous mentor

okuzan (approved):

PR #1111, opened 2025-01-20
initial solution

VitaliyJvM:


src/main/java/core/basesyntax/exception/InvalidDataException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.exception;
+
+public class InvalidDataException extends Exception {
```
Elena-Bruyako, line 3:
```suggestion
public class RegistrationException extends Exception {
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,32 @@
...
+                    + MIN_LOGIN_LENGTH + " characters!");
+        }
+        if (user.getPassword() == null || user.getPassword().length() < MIN_PASSWORD_LENGTH) {
+            throw new InvalidDataException("User's password  has to be at least "
+                    + MIN_PASSWORD_LENGTH + " characters!");
+        }
+        if (user.getAge() == null || user.getAge() < MIN_AGE) {
+            throw new InvalidDataException("User's age should be at least "
+                    + MIN_AGE + " years old!");
+        }
```
Elena-Bruyako, line 30: What if user exists with the same login?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @BeforeEach
+    void setUp() {
+        testUser.setLogin("Albertino");
+        testUser.setPassword("password");
+        testUser.setAge(25);
+    }
+
+    @Test
+    void nullUser_NotOk() {
+        try {
```
Elena-Bruyako, line 31: remove
just use assertThrows

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Test
+    void correctPasswordLength_Ok() {
+        testUser.setPassword("Bobina");
+        try {
+            User registeredUser = registrationService.register(testUser);
+            int actual = registeredUser.getPassword().length();
+            assertEquals(6, actual);
+        } catch (InvalidDataException e) {
+            throw new RuntimeException(e);
+        }
```
Elena-Bruyako, line 80:
```suggestion
            User registeredUser = registrationService.register(testUser);
            assertEquals(registeredUser, testUser);
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,82 @@
...
+        testUser.setPassword(null);
+        assertThrows(InvalidDataException.class, () -> registrationService.register(testUser),
+                errorMessage);
+        testUser.setPassword("Bob");
+        assertThrows(InvalidDataException.class, () -> registrationService.register(testUser),
+                errorMessage);
+    }
+
+    @Test
+    void correctPasswordLength_Ok() {
```
Elena-Bruyako, line 72: check task description and use particular naming pattern for test methods

```suggestion
    void validUser_Ok() {
```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,32 @@
...
+        if (user.getPassword() == null || user.getPassword().length() < MIN_PASSWORD_LENGTH) {
+            throw new RegistrationException("User's password  has to be at least "
+                    + MIN_PASSWORD_LENGTH + " characters!");
+        }
+        if (user.getAge() == null || user.getAge() < MIN_AGE) {
+            throw new RegistrationException("User's age should be at least "
+                    + MIN_AGE + " years old!");
+        }
+        return (storageDao.get(user.getLogin()) == null) ? storageDao.add(user) : user;
```
JJJazl, line 31: Your version is obviously short, but this version of the code is harder to debug and impossible to log, so let's add a few extra lines and make your code easier to read, understand and maintain + let's check if the user already exists and add a test for that.
```suggestion
        if (storageDao.contains(user.getLogin)) {
            throw RegistrationException()
        }
        storageDao.add(user);
        return user;
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,72 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.exception.RegistrationException;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
```
JJJazl, line 12: Add test to check if the user already exists in storage

JJJazl (changes requested):

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
VitaliyJvM: I have not implemented this functional only because of "Make sure to not add something that is not described in the task requirements, or tied to something that may change over time like max user age."
And I think "storageDao.contains(user.getLogin)" won't work because our storage stores User objects
;)

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+    private static RegistrationService registrationService;
+    private final User testUser = new User();
+    private final StorageDao storageDao = new StorageDaoImpl();
+
+    @BeforeAll
+    static void beforeAll() {
+        registrationService = new RegistrationServiceImpl();
+    }
```
ahoienko, line 23: What is the reason to use BeforeAll here? All tests should be independent and should not use shared objects, it means that before each test we should create new object that we are testing

Why some dependencies initialized in fields and some in beforeAll?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+import core.basesyntax.dao.StorageDaoImpl;
+import core.basesyntax.exception.RegistrationException;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
+    private static RegistrationService registrationService;
+    private final User testUser = new User();
```
ahoienko, line 17: Better to create new user in each test, in this case it wont be situation when somebody changed the user and forgot to override some user fields

ahoienko (changes requested):
Nice, left some comments

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
VitaliyJvM: I make a testUser once and initialize all its fields in setUp() method (BeforeEach tests).

VitaliyJvM: For this purpose my setUp() method renew all fields of the testUser  @BeforeEach tests

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Test
+    void registerUser_validPassword_Ok() throws RegistrationException {
+        testUser.setPassword("Bobina");
+        User registeredUser = registrationService.register(testUser);
+        assertEquals(registeredUser, testUser);
+    }
+
+    @Test
+    void registerUser_userExists_Ok() throws RegistrationException {
```
liliia-ponomarenko, line 80:
```suggestion
    void registerUser_validData_Ok() throws RegistrationException {

```

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
liliia-ponomarenko: Not fixed. You should throw an exception if a user with the same login is already registered. Don’t  forget to test it

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Test
+    void registerUser_shortPassword_NotOk() {
+        String errorMessage = "User's password is less than 6 symbols";
+        testUser.setPassword(null);
+        assertThrows(RegistrationException.class, () -> registrationService.register(testUser),
+                errorMessage);
+        testUser.setPassword("Bob");
+        assertThrows(RegistrationException.class, () -> registrationService.register(testUser),
+                errorMessage);
```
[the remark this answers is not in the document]
VitaliyJvM: There are some tips in "checklist.md" file like: "Few not-ok checks you can have in one not-ok test method. Same for ok-test."
Don't this suit it?

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
VitaliyJvM: Here I try to check whether user exists on request of @JJJazl : "+ let's check if the user already exists and add a test for that."
That's why I call this method in this way

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,35 @@
...
 public class RegistrationServiceImpl implements RegistrationService {
+    private static final int MIN_AGE = 18;
+    private static final int MIN_LOGIN_LENGTH = 6;
+    private static final int MIN_PASSWORD_LENGTH = 6;
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
-    public User register(User user) {
-        return null;
+    public User register(User user) throws RegistrationException {
```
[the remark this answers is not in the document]
VitaliyJvM: There is a task in "Readme.me" :
"You should create your custom unchecked exception and throw it in case of invalid data."

In tutorial video "JUnit best practices" StringUtil class has method "toUpperCase" that throws exception in this way.

So I can't understand your prompt.

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
Elena-Bruyako: unit test should check particular method - register in this case, so this test case redundant and you can remove it

 @JJJazl : "+ let's check if the user already exists and add a test for that." 
meant this test case `registerUser_twiceRegister_NotOk`

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @Test
+    void registerUser_validPassword_Ok() throws RegistrationException {
+        testUser.setPassword("Bobina");
+        User registeredUser = registrationService.register(testUser);
+        assertEquals(registeredUser, testUser);
+    }
+
+    @Test
+    void registerUser_twiceRegister_NotOk() throws RegistrationException {
+        registrationService.register(testUser);
```
Elena-Bruyako, line 86:
```suggestion
        Storage.people.add(testUser);
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+    void registerUser_shortPassword_NotOk() {
+        String errorMessage = "User's password is less than 6 symbols";
+        testUser.setPassword("Bob");
+        assertThrows(RegistrationException.class, () -> registrationService.register(testUser),
+                errorMessage);
+    }
+
+    @Test
+    void registerUser_validPassword_Ok() throws RegistrationException {
+        testUser.setPassword("Bobina");
```
Elena-Bruyako, line 79:
```suggestion
    void registerUser_validUser_Ok() throws RegistrationException {
```

Elena-Bruyako (approved):

PR #1163, opened 2025-05-14
implemented register() method and added tests

DZotsenko:


src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,193 @@
...
+        assertEquals(eightCharacterLoginUser, service.register(eightCharacterLoginUser));
+    }
+
+    @Test
+    void register_loginAlreadyExists_notOk() {
+        User loginAlreadyExistsUser = new User();
+        loginAlreadyExistsUser.setLogin("dianazot");
+        loginAlreadyExistsUser.setPassword("testpassword");
+        loginAlreadyExistsUser.setAge(19);
+        assertThrows(RegistrationException.class, () -> service.register(loginAlreadyExistsUser));
```
sarakhmen: the user isn't manually added in this test but the exception is thrown. It means that results from the previous tests affect this one.
Let's add @aftereach method that will clear our storage of users

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,193 @@
...
+        assertThrows(RegistrationException.class, () -> service.register(threeCharacterLoginUser));
+    }
+
+    @Test
+    void register_fiveCharacterLogin_notOk() {
+        User fivecharacterLoginUser = new User();
+        fivecharacterLoginUser.setLogin("diana");
+        fivecharacterLoginUser.setPassword("testpassword");
+        fivecharacterLoginUser.setAge(19);
+        assertThrows(RegistrationException.class, () -> service.register(fivecharacterLoginUser));
```
sarakhmen: let's also validate error messages in each test where the exception is thrown

sarakhmen (changes requested):
Let's improve your solution :)

Elena-Bruyako (approved):

PR #1174, opened 2025-06-11
test commit

kravchenko1207:


liliia-ponomarenko (changes requested):
Let's implement your HW

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,37 @@
...
+        if (user.getAge() == null || user.getAge() < 18) {
+            throw new RegistrationException("User age must be at least 18 years old");
+        }
+
+        if (storageDao.get(user.getLogin()) != null) {
+            throw new RegistrationException("User with thid login already exists");
+        }
+
+        return storageDao.add(user);
+
```
sarakhmen, line 36:
```suggestion
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,109 @@
...
+        user1.setPassword("123456");
+        user1.setAge(19);
+        registrationService.register(user1);
+
+        User user2 = new User();
+        user2.setLogin("userSameLogin");
+        user2.setPassword("123456");
+        user2.setAge(20);
+        assertThrows(RegistrationException.class, () -> registrationService.register(user2));
+    }
```
sarakhmen: let's not use our service twice to simulate existing user as in this case our test is not isolated and independent enough. Please add user directly to the storage using StorageDao instead

sarakhmen (changes requested):
good job! just a minor improvement required

sarakhmen (approved):

PR #1175, opened 2025-06-19
first solution

anpiet:


src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -2,13 +2,29 @@
...
 public class RegistrationServiceImpl implements RegistrationService {
     private final StorageDao storageDao = new StorageDaoImpl();
 
     @Override
     public User register(User user) {
-        return null;
+        if (user == null) {
+            throw new RegistrationException("User is null");
+        }
+        if (storageDao.get(user.getLogin()) != null) {
```
slade13, line 16: You check “duplicate login” before “login length.” That’s OK, but if someone passes a 3‑char login that matches an existing user’s short login, you’ll throw “User already exists” rather than “Login too short.” Consider checking null/length first, then uniqueness, so users always see the most relevant error

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import core.basesyntax.exception.RegistrationException;
+import core.basesyntax.model.User;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class RegistrationServiceImplTest {
```
slade13, line 12: Duplicate‑login case you need a test like:
```
@Test
void register_duplicateLogin_notOk() {
  User u = new User("dupLogin", "goodPwd", 30);
  registrationService.register(u);   // first time OK
  assertThrows(RegistrationException.class,
    () -> registrationService.register(u),
    "Should reject duplicate login");
}
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+    private RegistrationService registrationService;
```
slade13, line 13: Add edge‑boundary success cases:

1. Login exactly 6 chars: should pass
2. Password exactly 6 chars: should pass
3. Age exactly 18: should pass

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Test
+    public void register_shortPassword_NotOk() {
+        User user = new User();
+        user.setLogin("validLogin");
+        user.setPassword("short");
+        user.setAge(20);
+
+        assertThrows(RegistrationException.class, () -> registrationService.register(user));
```
slade13, line 84: You currently only assert that an exception is thrown. To make sure your messages stay correct, you can capture and assert the message:
```
RegistrationException ex = assertThrows(...);
assertEquals("Login must be at least 6 characters long", ex.getMessage());
```

src/test/java/core/basesyntax/service/RegistrationServiceImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+        user.setPassword("validPassword");
+        user.setAge(20);
+
+        User actual = registrationService.register(user);
+        assertNotNull(actual);
+        assertEquals(user.getLogin(), actual.getLogin());
+    }
+
+    @Test
+    public void register_nullUser_NotOk() {
```
slade13, line 33: Your test names use `register_nullUser_NotOk`, etc. That’s consistent, but mixing upper‑ and lower‑case `Ok`/`NotOk` can read inconsistently. You may prefer:
```
register_nullUser_notOk
register_validUser_ok
register_shortLogin_notOk
```
to keep `<state>_<expectedBehavior>` fully lowercase for the expected behavior.

slade13 (approved):
Good job! Please review my comments and fix your code accordingly.

PR #1248, opened 2025-12-09
Hw registration

sannchous:


sannchous:
I believe that the method for checking characters does not interfere with the operation of the program, with this method I added a little "my own" to the code, so I am against removing this method.

Elena-Bruyako (approved):

PR #1313, opened 2026-06-11
Created tests for service registrationService

Inna1989F:
I added tests


src/main/java/core/basesyntax/model/User.java
```diff
@@ -8,6 +8,13 @@ public class User {
     private String password;
     private Integer age;
 
+    public User(Long id, String login, String password,Integer age) {
```
Elena-Bruyako, line 11: don't provide Id to constructor

src/main/java/core/basesyntax/service/RegistrationServiceImpl.java
```diff
@@ -1,14 +1,48 @@
...
 
 public class RegistrationServiceImpl implements RegistrationService {
-    private final StorageDao storageDao = new StorageDaoImpl();
+    private static final int MIN_LOGIN_LENGTH = 6;
+    private static final int MIN_PASSWORD_LENGTH = 6;
+    private static final int MIN_AGE = 18;
+
+    private final StorageDao storageDao;
+
+    public RegistrationServiceImpl(StorageDao storageDao) {
```
Elena-Bruyako, line 14: Why do you need this constructor?

Elena-Bruyako (changes requested):
see comments

